PR #1, opened 2015-11-06
Improve user ergonomics, performance and deny a timing attack

Ryman:
(sorry for the bundled commits, I can separate them into separate PRs if you need)

This looks pretty useful so I wanted to help improve the ergonomics a bit (and then got a bit sidetracked in making it faster). Most of the ownership was not required (so `to_owned` is generally unnecessary), as the majority of work cannot be done in-place (at least not with current libraries).

This commit fixes a potential timing attack on hmac verification: 486c4a87f9603250167e40d70da2845e04a07500

Performance change:
```bash
# before
test bench_decode ... bench:      11,501 ns/iter (+/- 885)
test bench_encode ... bench:       5,036 ns/iter (+/- 511)

# after
test bench_decode ... bench:       3,460 ns/iter (+/- 718)
test bench_encode ... bench:       3,964 ns/iter (+/- 516)
```

### DIFF
--- a/examples/claims.rs
+++ b/examples/claims.rs
@@ -16,7 +16,7 @@ fn main() {
         company: "ACME".to_owned()
     };
     let key = "secret";
-    let token = match encode::<Claims>(my_claims, key.to_owned(), Algorithm::HS256) {
+    let token = match encode(&my_claims, key, Algorithm::HS256) {
         Ok(t) => t,
         Err(_) => panic!() // in practice you would return the error
     };
--- a/examples/claims.rs
+++ b/examples/claims.rs
@@ -21,7 +21,7 @@ fn main() {
         Err(_) => panic!() // in practice you would return the error
     };
 
-    let claims = match decode::<Claims>(token.to_owned(), key.to_owned(), Algorithm::HS256) {
+    let claims = match decode::<Claims>(&token, key, Algorithm::HS256) {
         Ok(c) => c,
         Err(err) => match err {
             Error::InvalidToken => panic!(), // Example on how to handle a specific error
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -159,7 +159,7 @@ mod tests {
 
     #[test]
     fn sign_hs256() {
-        let result = sign("hello world", "secret".as_bytes(), Algorithm::HS256);
+        let result = sign("hello world", b"secret", Algorithm::HS256);
         let expected = "c0zGLzKEFWj0VxWuufTXiRMk5tlI5MbGDAYhzaxIYjo";
         assert_eq!(result, expected);
     }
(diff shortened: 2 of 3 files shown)


Keats:
Don't worry for the separated commits, going to have a look now

examples/claims.rs
```diff
@@ -16,12 +16,12 @@ fn main() {
         company: "ACME".to_owned()
     };
     let key = "secret";
-    let token = match encode::<Claims>(my_claims, key.to_owned(), Algorithm::HS256) {
+    let token = match encode(&my_claims, key, Algorithm::HS256) {
```
Keats: :+1:

src/lib.rs
```diff
@@ -29,40 +30,69 @@ pub enum Algorithm {
 /// A part of the JWT: header and claims specifically
 /// Allows converting from/to struct with base64
 pub trait Part {
-    fn from_base64(encoded: String) -> Result<Self, Error> where Self: Sized;
-    fn to_base64(&self) -> Result<String, Error>;
+    type Encoded: AsRef<str>;
```
Keats: just read the AsRef entry, that's nice

src/lib.rs
```diff
@@ -29,40 +30,69 @@ pub enum Algorithm {
...
+    fn to_base64(&self) -> Result<Self::Encoded, Error> {
+        let encoded = match self.alg {
+            Algorithm::HS256 => { "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9" },
+            Algorithm::HS384 => { "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzM4NCJ9" },
+            Algorithm::HS512 => { "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzUxMiJ9" },
+        };
+
+        Ok(encoded)
+    }
+}
```
Keats: The issue with that is that you loose any kind of support for other parameters in the header.
For example, google oauth seems to use the `kid` one (http://stackoverflow.com/questions/20159782/how-can-i-decode-a-google-oauth-2-0-jwt-openid-connect-in-a-node-app).

src/lib.rs
```diff
@@ -114,14 +151,14 @@ pub fn decode<T: Part>(token: String, secret: String, algorithm: Algorithm) -> R
         return Err(Error::InvalidSignature);
     }
 
-    // not reachable right now
-    let header = try!(Header::from_base64(parts[0].to_owned()));
+    let (token, header) = expect_two!(payload.rsplitn(2, '.'));
```
Keats: maybe token -> claims?

Keats:
Much nicer overall!

src/lib.rs
Ryman: I think it can work in tandem, if `kid` or something dynamic is used then do the computation, else use the "known" values. (the `Encoded` assoc-type could be [`Cow<'static, str>`](http://doc.rust-lang.org/std/borrow/enum.Cow.html))

src/lib.rs
Ryman: Renamed and rebased

Keats:
Cheers!

PR #3, opened 2015-12-13
Allow non-UTF8 keys for decoding too

### LINKED ISSUE #2: `decode` doesn't take arbitrary bytes as input
I have a key that is not valid UTF-8. Currently it's possible to encode, but not decode using such a key.

I've been working on a PR that makes `decode` take `AsRef<[u8]>` (like `encode`), but on the way I've found that this breaks `decode::<Claims>(..)` usage because of additional type parameters. The other options are:

* `decode<T: Part>(token: &str, secret: &AsRef<[u8]>, algorithm: ...) -> Result<T, ...>`, no idea why this works???

* Just advise users to use different forms of typehinting if possible, and only use `decode::<Claims, _>` if necessary.

untitaker:
Fix #2

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -127,7 +127,15 @@ pub fn encode<T: Part, B: AsRef<[u8]>>(claims: &T, secret: B, algorithm: Algorit
 
 /// Decode a token into a Claims struct
 /// If the token or its signature is invalid, it will return an error
-pub fn decode<T: Part>(token: &str, secret: &str, algorithm: Algorithm) -> Result<T, Error> {
+pub fn decode<T: Part>(token: &str, secret: &[u8], algorithm: Algorithm) -> Result<T, Error> {
+    // We don't use AsRef<[u8]> for `secret` because it would require changing this:
+    //
+    //     decode::<MyStruct>(...)
+    //
+    // to:
+    //
+    //     decode::<MyStruct, _>(...)
+
     macro_rules! expect_two {
         ($iter:expr) => {{
             let mut i = $iter; // evaluate the expr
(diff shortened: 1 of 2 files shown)


untitaker:
I just got word from somebody else (@xmppwocky) that this *should* work.

untitaker:
Eek, that doesn't even work with &str!

untitaker:
This leaves us with two options:

* `decode::<MyClaims, _>(token, "secret")`
* `decode::<MyClaims>(token, "secret".as_ref())`

For the sake of backwards compat you could also have a dedicated function `decode2` that has either of those interfaces, and make `decode` wrap around it.

Keats:
Ha too bad.
I don't have strong feelings for any of the choice to be honest but i think the second one (`as_ref`) might be easier to figure out what you forgot it from the compiler error so i'd lean towards that.

I will add a changelog section in the readme to document that, no need for backwards compat yet imo

untitaker:
@Keats Updated, please take a look.

src/lib.rs
```diff
@@ -127,7 +127,15 @@ pub fn encode<T: Part, B: AsRef<[u8]>>(claims: &T, secret: B, algorithm: Algorit
...
 /// If the token or its signature is invalid, it will return an error
-pub fn decode<T: Part>(token: &str, secret: &str, algorithm: Algorithm) -> Result<T, Error> {
+pub fn decode<T: Part>(token: &str, secret: &[u8], algorithm: Algorithm) -> Result<T, Error> {
+    // We use &AsRef<[u8]> instead of a separate type parameter such that one still can use:
+    //
+    //     decode::<MyStruct>(...)
+    //
+    // instead of:
+    //
+    //     decode::<MyStruct, _>(...)
```
Keats: that comment is not valid anymore is it?

untitaker: Updated.

Keats:
Thanks!

PR #5, opened 2015-12-17
Use normal json decoding for header

### LINKED ISSUE #4: Header::from_base64 doesn't accept/"just discard" extra fields or differently ordered JSON
This header works:

    {
      "alg": "HS256",
      "typ": "JWT"
    }

This doesn't:

    {
      "typ": "JWT",
      "alg": "HS256"
    }

untitaker:
Fix #4

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,7 +8,7 @@
 extern crate rustc_serialize;
 extern crate crypto;
 
-use rustc_serialize::{json, Encodable, Decodable};
+use rustc_serialize::{json, Encoder, Decoder, Encodable, Decodable};
 use rustc_serialize::base64::{self, ToBase64, FromBase64};
 use crypto::sha2::{Sha256, Sha384, Sha512};
 use crypto::hmac::Hmac;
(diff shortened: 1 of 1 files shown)


Keats:
I was more thinking on hardcoding the 3 other orders for now as I'm not a big fan of discarding things silently.
Imo it's either the "cheap" way (harcoding the headers if it's only <10 like now) or allowing custom headers completely

untitaker:
How would you implement custom headers? I thought of having a new type param `H: Part` for `encode` and `decode`

Keats:
2 ways I can see:

- 1 method for encode/decode with a custom header in addition to the current on
- 1 method for both: instead of passing the algorithm when encoding you would pass a Header obj

If we follow the spec, we can do the following:
- add some Option to header for the ones referenced in http://self-issued.info/docs/draft-jones-json-web-token-01.html#ReservedHeaderParameterName 
- add a Default Header with alg=HS256 and everything else to None
- keep the constructor with the alg as param if we just want to create a basic header with another alg
- if you want to add some other params though you do it by hand after instantiation
- decode should still work the same

what do you think ?

untitaker:
What if the user doesn't know e.g. the value of `cty` before decoding?

Keats:
we'd need to return the header in decode as well, forgot to mention that.

`cty` seems to only be used when when the token is encrypted, which this crate doesn't support

untitaker:
See #6

PR #6, opened 2015-12-18
Add all params of header from RFC

Keats:


Keats:
@untitaker that's what I had in mind but i'm not too happy about the `TokenData` struct existence.

It's also twice slower for encoding and 4x slower for decoding

Keats:
(that's still fast enough mind you)

src/lib.rs
```diff
@@ -51,46 +51,41 @@ impl<T> Part for T where T: Encodable + Decodable {
...
-    fn to_base64(&self) -> Result<Self::Encoded, Error> {
-        let encoded = match self.alg {
-            Algorithm::HS256 => { "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9" },
-            Algorithm::HS384 => { "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzM4NCJ9" },
-            Algorithm::HS512 => { "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzUxMiJ9" },
-        };
-
-        Ok(encoded)
-    }
+#[derive(Debug)]
```
untitaker: You could replace this type with `(Header, T)`

untitaker:
The perf results are confusing. Shouldn't signature verification be the bottleneck?

src/lib.rs
untitaker: On second thought, all fields of `Header` could be moved to `TokenData`.

Keats: what do you mean?

Keats:
I will have a look at the perfs when I have time as I'm a bit curious as well

src/lib.rs
untitaker: struct TokenData<T: Part> {
        claims: T,
        typ: String,
        alg: Algorithm,
        jku: Option<String>,
        kid: Option<String>,
        x5u: Option<String>,
        x5t: Option<String>
    }

Keats: This blurs the line between header and claims a bit too much for my taste. I've looked at tuple but it seems like it would be a downgrade compared to a struct in this case

Keats:
I ran https://github.com/erickt/rust-serialization-benchmarks for rustc_serialize/serde and rustc_serialize seems quite slow at decoding so that must be it, as the only change in decode is going from pattern matching to json + base64 decoding

untitaker:
:+1: on the architecture. Bikeshedding:

* TokenData could be named just `Token`. This probably avoids the interpretation of the "data"-part being only the JWT claims
* decode/encode could be methods on that struct

Keats:
I thought of calling it `Token` initially but there might be some confusion on the name, ie even in my example and my code I call the string `token`, not that claims/header. So in practice you might already have a `token` variable (as a param in the function ) and you would need to find a new name for the `decode` result, which is `TokenData` here. Open to better suggestions though

Second point is interesting, I'll play a bit with it and see how the API feels like when I have time

Keats:
So for the second point the api would look something like:

```rust
use jwt::{Token, decode, Header, Algorithm};

#[derive(Debug, RustcEncodable, RustcDecodable)]
struct Claims {
    sub: String,
    company: String
}

let my_claims = Claims {
     sub: "b@b.com".to_owned(),
     company: "ACME".to_owned()
};

/// encode default header
let token = Token { claims: my_claims, header: Header::default()};
return token.encode("blabla".as_ref());

/// encode custom header
let mut header = Header::default();
header.kid = Some("signing_key".to_owned());
let token = Token { claims: my_claims, header: header};
return token.encode("blabla".as_ref());

/// decode
let token = decode::<Claims>(&token_str, "blabla".as_ref(), Algorithm::HS256)
```
It's nice but doesn't feel like a significant improvement so I'm not sure

untitaker:
lgtm either way

Keats:
Pushed 1.0.0 on crates.io

PR #10, opened 2016-03-19
custom json encoder omits None when serializing

WartsWerdna:
Hi, this pull request prevents optional header fields with value `None` from being serialized and therefore saves a few bytes in the resulting token. This is done by creating an implementation of the `rustc_serialize::Encodable` trait (instead of using automatically deriving `RustcEncodable`).


Keats:
Great, thanks!

PR #14, opened 2016-08-16
Repitition in optional! to further reduce boilerplate

crossroads1112:
Using repitition allows you to use `optional!` for all struct fields at once.


Keats:
it doesn't compile according to travis, i like the idea though

PR #26, opened 2017-04-01
updated ring dependency to 0.7

celaus:
Hi, I saw your plans for v2, but I needed the ring dependency to be updated sooner ;)


Keats:
Pushed as 1.1.7, thanks!

Issue #27, opened 2017-04-21
Update to serde 1.0


Keats:
That requires changing the `serialization` file as well for lifetimes

Keats:
@mike-engel if you have time and want to do it, you can pick this one up!

mike-engel:
@Keats I'll take a stab at it today/this weekend 😄

PR #28, opened 2017-04-21
Update Serde to 1.0

mike-engel:
This is PR updates serde to 1.0. The README and Cargo file have been updated to use 1.0. This will close out #27.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,7 +31,7 @@ pub use crypto::{
 pub use validation::Validation;
 
 
-use serde::de::Deserialize;
+use serde::de::DeserializeOwned;
 use serde::ser::Serialize;
 
 use errors::{Result, ErrorKind};
@@ -101,7 +101,7 @@ macro_rules! expect_two {
 /// // Claims is a struct that implements Deserialize
 /// let token_data = decode::<Claims>(&token, "secret", Algorithm::HS256, &Validation::default());
 /// ```
-pub fn decode<T: Deserialize>(token: &str, key: &[u8], algorithm: Algorithm, validation: &Validation) -> Result<TokenData<T>> {
+pub fn decode<T: DeserializeOwned>(token: &str, key: &[u8], algorithm: Algorithm, validation: &Validation) -> Result<TokenData<T>> {
     let (signature, signing_input) = expect_two!(token.rsplitn(2, '.'));
 
     if validation.validate_signature && !verify(signature, signing_input, key, algorithm)? {
--- a/src/serialization.rs
+++ b/src/serialization.rs
@@ -1,5 +1,5 @@
 use base64;
-use serde::de::Deserialize;
+use serde::de::DeserializeOwned;
 use serde::ser::Serialize;
 use serde_json::{from_str, to_string, Value};
 use serde_json::map::Map;
@@ -10,7 +10,7 @@ use header::Header;
 
 /// The return type of a successful call to decode
 #[derive(Debug)]
-pub struct TokenData<T: Deserialize> {
+pub struct TokenData<T: DeserializeOwned> {
     pub header: Header,
     pub claims: T
 }
@@ -22,7 +22,7 @@ pub fn to_jwt_part<T: Serialize>(input: &T) -> Result<String> {
 }
 
 /// Decodes from base64 and deserializes from JSON to a struct
-pub fn from_jwt_part<B: AsRef<str>, T: Deserialize>(encoded: B) -> Result<T> {
+pub fn from_jwt_part<B: AsRef<str>, T: DeserializeOwned>(encoded: B) -> Result<T> {
     let decoded = base64::decode_config(encoded.as_ref(), base64::URL_SAFE_NO_PAD)?;
     let s = String::from_utf8(decoded)?;
 
@@ -30,7 +30,7 @@ pub fn from_jwt_part<B: AsRef<str>, T: Deserialize>(encoded: B) -> Result<T> {
 }
 
 /// Decodes from base64 and deserializes from JSON to a struct AND a hashmap
-pub fn from_jwt_part_claims<B: AsRef<str>, T: Deserialize>(encoded: B) -> Result<(T, Map<String, Value>)> {
+pub fn from_jwt_part_claims<B: AsRef<str>, T: DeserializeOwned>(encoded: B) -> Result<(T, Map<String, Value>)> {
     let decoded = base64::decode_config(encoded.as_ref(), base64::URL_SAFE_NO_PAD)?;
     let s = String::from_utf8(decoded)?;
 
--- a/src/serialization.rs
+++ b/src/serialization.rs
@@ -10,7 +10,7 @@ use header::Header;
 
 /// The return type of a successful call to decode
 #[derive(Debug)]
-pub struct TokenData<T: DeserializeOwned> {
+pub struct TokenData<T> {
     pub header: Header,
     pub claims: T
 }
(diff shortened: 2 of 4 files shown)


mike-engel:
As of right now, there are two lifetime issues [here](https://github.com/mike-engel/rust-jwt/blob/serde-1.0/src/serialization.rs#L31) and [here](https://github.com/mike-engel/rust-jwt/blob/serde-1.0/src/serialization.rs#L41). I've tried using `str::from_utf8`, but that's no good either since then `decoded` goes out of scope too quick.

@Keats I'll keep trying here, but I don't think I have enough Rust experience with how to combat these two lifetime issues. If you have any ideas, let me know!

```
error: `s` does not live long enough
  --> src/serialization.rs:31:18
   |
31 |     Ok(from_str(&s)?)
   |                  ^ does not live long enough
32 | }
   | - borrowed value only lives until here
   |
note: borrowed value must be valid for the lifetime 'a as defined on the body at 27:85...
  --> src/serialization.rs:27:86
   |
27 |   pub fn from_jwt_part<'a, B: AsRef<str>, T: Deserialize<'a>>(encoded: B) -> Result<T> {
   |  ______________________________________________________________________________________^ starting here...
28 | |     let decoded = base64::decode_config(encoded.as_ref(), base64::URL_SAFE_NO_PAD)?;
29 | |     let s = String::from_utf8(decoded)?;
30 | |
31 | |     Ok(from_str(&s)?)
32 | | }
   | |_^ ...ending here

error: `s` does not live long enough
  --> src/serialization.rs:41:31
   |
41 |     let claims: T = from_str(&s)?;
   |                               ^ does not live long enough
...
44 | }
   | - borrowed value only lives until here
   |
note: borrowed value must be valid for the lifetime 'a as defined on the body at 37:40...
  --> src/serialization.rs:37:41
   |
37 |        -> Result<(T, Map<String, Value>)> {
   |  _________________________________________^ starting here...
38 | |     let decoded = base64::decode_config(encoded.as_ref(), base64::URL_SAFE_NO_PAD)?;
39 | |     let s = String::from_utf8(decoded)?;
40 | |
41 | |     let claims: T = from_str(&s)?;
42 | |     let map: Map<_, _> = from_str(&s)?;
43 | |     Ok((claims, map))
44 | | }
   | |_^ ...ending here
```

Keats:
Thanks!
Actually I just noticed there is a https://docs.serde.rs/serde/de/trait.DeserializeOwned.html trait that might be more what we want since as a library we don't really know how the end struct is going to be used.

Is that how it's meant to be used @dtolnay (sorry for the random ping)?

dtolnay:
I always appreciate a ping! Yes, both `from_jwt_part` and `from_jwt_part_claims` require `T: DeserializeOwned` because they are deserializing from the base64-decoded data so they shouldn't be allowed to borrow anything (the base64-decoded goes out of scope when the function returns).

I will leave a few review comments as well.

src/lib.rs
```diff
@@ -101,7 +97,13 @@ macro_rules! expect_two {
 /// // Claims is a struct that implements Deserialize
 /// let token_data = decode::<Claims>(&token, "secret", Algorithm::HS256, &Validation::default());
 /// ```
-pub fn decode<T: Deserialize>(token: &str, key: &[u8], algorithm: Algorithm, validation: &Validation) -> Result<TokenData<T>> {
+pub fn decode<T>(token: &str,
+                 key: &[u8],
+                 algorithm: Algorithm,
+                 validation: &Validation)
+                 -> Result<TokenData<T>>
+    where for<'a> T: Deserialize<'a>
```
dtolnay: This would be more readable as `T: DeserializeOwned`. This function requires that `T` can deserialize without borrowing any data because the string it is deserialized from is thrown away before this function returns.

src/serialization.rs
```diff
@@ -4,15 +4,17 @@ use serde::ser::Serialize;
...
-use errors::{Result};
+use errors::Result;
 use header::Header;
 
 
 /// The return type of a successful call to decode
 #[derive(Debug)]
-pub struct TokenData<T: Deserialize> {
+pub struct TokenData<T>
+    where for<'a> T: Deserialize<'a>
```
dtolnay: This bound should not be here. Structs in Rust should never have trait bounds. Trait bounds are for behavior.

```rust
impl<T> Data<T> where T: Behavior { /* behavior, okay */ }

fn behave<T>() where T: Behavior { /* also behavior, also okay */ }

struct Data<T> where T: Behavior { /* don't do this */ }
```

The only exception is things like [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html) that use associated types to define data.

src/crypto.rs
```diff
@@ -29,10 +29,8 @@ pub enum Algorithm {
 /// The actual HS signing + encoding
 fn sign_hmac(alg: &'static digest::Algorithm, key: &[u8], signing_input: &str) -> Result<String> {
     let signing_key = hmac::SigningKey::new(alg, key);
-    Ok(base64::encode_config(
-        hmac::sign(&signing_key, signing_input.as_bytes()).as_ref(),
-        base64::URL_SAFE_NO_PAD
-    ))
+    Ok(base64::encode_config(hmac::sign(&signing_key, signing_input.as_bytes()).as_ref(),
```
dtolnay: Next time please make whitespace changes in a separate commit. They makes it hard to find and review the thing you are actually changing.

Keats: Agreed on that, @mike-engel can you revert all the formatting changes for now?
Is it rustfmt using the latest config? I thought it would be more block indented than that

dtolnay: Block style is currently recommended by the fmt-rfc folks but not the default in rustfmt. [This is the settings they recommend.](https://github.com/serde-rs/serde/blob/v1.0.0/rustfmt.toml)

dtolnay: More official source I guess: [here](https://github.com/rust-lang-nursery/rustfmt/blob/master/rfc-rustfmt.toml).

mike-engel: Yeah, sorry about that! Turns out my rustfmt was old. Would you prefer me to avoid rustfmt altogether, or add a config like the two linked from serde & rustmft?

Keats: Avoid it for now please, we can run it later

src/serialization.rs
mike-engel: Thanks for the clarification. If it's not obvious, I'm new to Rust, and this result was me fighting with lifetime errors from the compiler 😞

dtolnay: I filed https://github.com/Manishearth/rust-clippy/issues/1689 to catch this case with a warning in Clippy. Apparently even the standard library does it wrong sometimes!

mike-engel:
Ok, updated now with `DeserializedOwned`, which made everything much better. Sorry I didn't see that in the docs before, and thanks to @dtolnay for the suggestions, clarifications, work the work on serde 😄 

Tests and benchmarks run fine. If there's anything else you'd like me to fix/edit/add, let me know.

src/serialization.rs
```diff
@@ -10,7 +10,7 @@ use header::Header;
 
 /// The return type of a successful call to decode
 #[derive(Debug)]
-pub struct TokenData<T: Deserialize> {
+pub struct TokenData<T: DeserializeOwned> {
```
dtolnay: This bound should be removed: just `pub struct TokenData<T> { ... }`. See https://github.com/Manishearth/rust-clippy/issues/1689.

Keats:
Thanks to both of you!

PR #29, opened 2017-04-22
No need to pass an algorithm to decode

Keats:


Keats:
@mike-engel I think with that v2 is ready to go

mike-engel:
Awesome! Thoughts on making HS256 the default? I'm going back and forth on it.

Keats:
Default should be not validating the algorithm imo

PR #30, opened 2017-05-09
Update *ring* to 0.9.4.

briansmith:
Before *ring* 0.9.3, it was possible to link multiple versions of *ring* into a program, e.g. if one version depended on *ring* 0.6 and another depended on *ring* 0.9. Unfortunately, this doesn't work, because the linker doesn't know to how to link *ring*'s C/asm code correctly in that kind of situation. *ring* 0.9.3 added a flag to its Cargo.toml to tell the Rust toolchain to stop allowing multiple versions of *ring* to be linked into the same program, to prevent any problems this may cause.

*ring* 0.9.4 was released with an update to make some crates easier to update to 0.9.x.

if you consider this a breaking change, let me know and I'll bump the version number to 3.0.0 instead of 2.0.1.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "jsonwebtoken"
-version = "2.0.0"
+version = "2.0.1"
 authors = ["Vincent Prouillet <vincent@wearewizards.io>"]
 license = "MIT"
 readme = "README.md"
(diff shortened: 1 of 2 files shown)


Keats:
Thanks!
That's an interesting thing to consider whether this is a breaking change or not... From what you said a program having 2 different versions wouldn't really work so in theory it shouldn't be a breaking change. I'll publish it as 2.0.1 and wait to see if anyone complains I guess 😛

briansmith:
Great! Thanks!

PR #33, opened 2017-06-23
Update dependencies

JacobChang:
If you think 2.0.2 is not an appropriate version, let me know and I'll change that


Keats:
Thanks, I'll check a bit more but i think the version number should be ok

PR #34, opened 2017-07-02
Make the `TokenData` struct public

mike-engel:
While upgrading from v1 to v2 of jsonwebtoken, the compiler was complaining that the `TokenData` struct was private, which indeed it it. It's marked `pub` within the `serialization` module, but that module isn't public itself.

I only made `TokenData` public, but could make the entire module public as well.

Happy to explain more or adjust as needed 😄

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target
 Cargo.lock
 .idea/
+.vscode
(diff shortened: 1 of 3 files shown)


Keats:
Thanks!

PR #35, opened 2017-07-03
Handle serde's f64 Number

mike-engel:
Serde 1.0, unlike Serde 0.9, only uses f64 for numbers. This was causing errors when decoding JWTs in serde was expecting a float but found an int for values like `iat`.

More info on a serde issue thread: https://github.com/serde-rs/json/issues/320


Keats:
Can you add a test that was failing before the change as well?

PR #39, opened 2017-08-25
V3

Keats:


Keats:
@naerbnic does this work for you for the `decode_header`?

@mhseiden @bhendo for the optional `typ`

Any other changes you'd like to see?

src/header.rs
```diff
@@ -8,7 +8,7 @@ pub struct Header {
     /// The type of JWS: it can only be "JWT" here
     ///
     /// Defined in [RFC7515#4.1.9](https://tools.ietf.org/html/rfc7515#section-4.1.9).
-    typ: String,
+    typ: Option<String>,
```
bhendo: @Keats Should this attribute be applied here? `#[serde(skip_serializing_if = "Option::is_none")]` ?

mhseiden: Fwiw we've needed to use that attribute in a lot of our GCP + serde code...

Keats: @bhendo yes it should, I just pushed  a patch

mhseiden:
The patch plays nice with our codebase; thanks for this!

bhendo:
@Keats looks like these changes do the trick for me. Thanks!

Keats:
@naerbnic ping if you want to have a look before a I merge it

Issue #41, opened 2017-10-18
Unsafe default behavior


idubrov:
If algorithm is not explicitly enforced via Validation, library will happily accept "forged" token where RS256 algorithm is replaced with HS256 and token is signed with RSA public key used as a secret.

Not sure what could be done here as this isn't the library issue per se, but rather unsafe usage of it. Maybe warning in the documentation?

Alternatively, library could accept the only one algorithm as a parameter. Since key you need to pass to decode is specific to the algorithm anyway, the current signature of (token: &str, key: &[u8], validation: &Validation) is not very useful.

[More details](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/)

idubrov:
Or, even better, instead of accepting key as a slice, it should accept closure which would return key based on provided algorithm type and "kid" (optional).

This would also avoid parsing header twice (since you already parsed it at the point you invoke "verify").

Keats:
Eh that's weird, I don't remember why Validation::algorithms is an `Option`, it shouldn't be.

Keats:
#42 removes the `Option` for algorithms so the end user will need to explicitely say which alg are allowed for starter

Keats:
I believe it's fixed now as the algorithm validation is now mandatory

PR #42, opened 2017-10-22
Make it mandatory to specify algorithm

Keats:


Keats:
@mike-engel @idubrov

Would that solve #41 and make it not too painful to use? It's annoying me a bit that I released 3.0 with optional alg validation :(

mike-engel:
@Keats I think it's a good solution. Yes, it's a little annoying, but in 99% of cases, you know the algorithm you want to validate, so it's not a big deal IMO. A couple extra lines of code.

PR #43, opened 2017-11-25
Fix jwt::decode documentation in readme

ChristophWurst:
Apparently jwt::decode expects the secret to be a slice. Thus blindly copying the example code from the readme gives a compilation error.

```
error[E0308]: mismatched types
  --> src/service/auth.rs:43:35
   |
43 |     jwt::decode::<Claims>(&token, "secret", &jwt::Validation::default())
   |                                   ^^^^^^^^ expected slice, found str
   |
   = note: expected type `&[u8]`
              found type `&'static str`
   = help: here are some functions which might fulfill your needs:
           - .as_bytes()
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Look at `examples/custom_header.rs` for a full working example.
 
 ### Decoding
 ```rust
-let token = decode::<Claims>(&token, "secret", &Validation::default())?;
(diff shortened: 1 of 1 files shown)


Keats:
Thanks!

PR #45, opened 2017-12-14
Implement "none" algorithm

etaoins:
This implements the "none" algorithm as defined in RFC 7519 section 6 ("Unsecured JWTs"). This algorithm has a dubious security history but it should be mitigated by two factors:

1) This crate requires explicitly specifying an algorithm for verification. Using this algorithm would require the caller to either specify it or write their own algorithm detection logic.

2) Sign/verify operations will fail if a key is specified. This would prevent misguided algorithm detection logic to use "none" with a key intended for RSA or HMAC.

A variant of this patch is being used internally by a proxy that converts signed JWTs to unsecured JWTs. This is part of a transition plan to use signed JWTs everywhere but this piece is needed for the migration.

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -11,6 +11,14 @@ use errors::{Result, ErrorKind};
 /// The algorithms supported for signing/verifying
 #[derive(Debug, PartialEq, Copy, Clone, Serialize, Deserialize)]
 pub enum Algorithm {
+    /// No signature
+    ///
+    /// This should be used with caution as it provides no guarantee of the authenticity of a
+    /// token. In particular this should not be selected as verification algorithm based on the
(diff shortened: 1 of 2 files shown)


Keats:
Not implementing the `none` algorithm was intentional as I couldn't find a usecase where one would use it. Let me think about this a bit more

etaoins:
Our legacy usecase was service-to-service identification on an private network. It's also required by RFC 7519:

```
Of the signature and MAC algorithms specified in JSON Web Algorithms
[JWA], only HMAC SHA-256 ("HS256") and "none" MUST be implemented by
conforming JWT implementations
```

The being said I understand why you'd be wary and we're happy to maintain a private fork.

PR #47, opened 2017-12-31
Add `serde` dependency to example Cargo.toml

chinedufn:
Hey @Keats ,

New to Rust so apologies if I missed  or overlooked anything obvious!

---

When trying to get started with your lib I got stuck on this compile time error:

image

After flailing for a bit I landed on a [serde GitHub issue that had a solution](https://github.com/serde-rs/serde/issues/856).

Notably:

> Adding serde to my dependencies fixes this.

---

So I went ahead and added `serde` to the example `Cargo.toml` in the README. Do let me know if there are any issues here!

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,9 @@
 Add the following to Cargo.toml:
 
 ```toml
-jsonwebtoken = "3"
+jsonwebtoken = "4"
 serde_derive = "1"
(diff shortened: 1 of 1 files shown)


Keats:
Woops, thanks for catching and fixing that! The CI failure is unrelated

Issue #48, opened 2018-02-24
Add a way to get the header + payload without verifying the signature


mike-engel:
Hey! I've gotten a new issue on my cli tool (https://github.com/mike-engel/jwt-cli/issues/4) which was fixed in 3.0 with `typ` being optional. 3.0 did, however, break one use case of my tool which is to just see the contents of a JWT without verifying the signature (debugging, etc). I used to use the `validate_signature` option in `Validation`, which was removed in 3.0.

Would you be open to having a separate method to just decode a token without verification? I looked through a couple of issues and didn't see the reason for removing `validate_signature`, but I'm sure there was one.

Thanks!

Keats:
The reason for removing it was to not have an insecure option as part of the main function. I think it would be fine to add something like `dangerous_unsafe_decode` to be added as an additional method, without touching the main `decode`.

Keats:
It's in 4.0.1

PR #49, opened 2018-03-03
Add a method to decode a token without signature validation

mike-engel:
Per #48, a new method should be added to decode a token without validating the signature. This is labelled as unsafe, since it definitely is. It solves my use case for `jwt-cli`, which the secret is generally not known.

- Solves #48
- `dangerous_unsafe_decode`
- No docs (aside from cargo) since people probably shouldn't use it

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -122,6 +122,35 @@ pub fn decode<T: DeserializeOwned>(token: &str, key: &[u8], validation: &Validat
     Ok(TokenData { header: header, claims: decoded_claims })
 }
 
+/// Decode a token without any signature validation into a struct containing 2 fields: `claims` and `header`.
+///
+/// NOTE: Do not use this unless you know what you are doing! If the token's signature is invalid, it will *not* return an error.
+///
+/// ```rust,ignore
+/// #[macro_use]
+/// extern crate serde_derive;
+/// use jsonwebtoken::{dangerous_unsafe_decode, Validation, Algorithm};
+///
+/// #[derive(Debug, Serialize, Deserialize)]
+/// struct Claims {
+///     sub: String,
+///     company: String
+/// }
+///
+/// let token = "a.jwt.token".to_string();
+/// // Claims is a struct that implements Deserialize
+/// let token_data = dangerous_unsafe_decode::<Claims>(&token, &Validation::new(Algorithm::HS256));
+/// ```
+pub fn dangerous_unsafe_decode<T: DeserializeOwned>(token: &str) -> Result<TokenData<T>> {
+    let (_, signing_input) = expect_two!(token.rsplitn(2, '.'));
+    let (claims, header) = expect_two!(signing_input.rsplitn(2, '.'));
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -122,6 +122,39 @@ pub fn decode<T: DeserializeOwned>(token: &str, key: &[u8], validation: &Validat
...
+/// struct Claims {
+///     sub: String,
+///     company: String
+/// }
+///
+/// let token = "a.jwt.token".to_string();
+/// // Claims is a struct that implements Deserialize
+/// let token_data = dangerous_unsafe_decode::<Claims>(&token, &Validation::new(Algorithm::HS256));
+/// ```
+pub fn dangerous_unsafe_decode<T: DeserializeOwned>(token: &str, validation: &Validation) -> Result<TokenData<T>> {
```
Keats: if you just want to decode it, do you even care about other validation? It could just take the `token`

src/lib.rs
```diff
@@ -122,6 +122,39 @@ pub fn decode<T: DeserializeOwned>(token: &str, key: &[u8], validation: &Validat
...
 }
 
+/// Decode a token without any signature validation into a struct containing 2 fields: `claims` and `header`.
+///
+/// NOTE: Do not use this unless you know what you are doing! If the token's signature is invalid, it will *not* return an error.
+///
+/// ```rust,ignore
+/// #[macro_use]
+/// extern crate serde_derive;
+/// use jsonwebtoken::{dangerous_unsave_decode, Validation, Algorithm};
```
Keats: unsave -> unsafe

src/lib.rs
```diff
@@ -122,6 +122,39 @@ pub fn decode<T: DeserializeOwned>(token: &str, key: &[u8], validation: &Validat
...
+/// let token_data = dangerous_unsafe_decode::<Claims>(&token, &Validation::new(Algorithm::HS256));
+/// ```
+pub fn dangerous_unsafe_decode<T: DeserializeOwned>(token: &str, validation: &Validation) -> Result<TokenData<T>> {
+    let (_, signing_input) = expect_two!(token.rsplitn(2, '.'));
+    let (claims, header) = expect_two!(signing_input.rsplitn(2, '.'));
+    let header: Header = from_jwt_part(header)?;
+
+    if !validation.algorithms.contains(&header.alg) {
+        return Err(ErrorKind::InvalidAlgorithm.into());
+    }
```
Keats: As mentioned above, if you're not checking the signature is it worth doing any kind of validation?

src/lib.rs
mike-engel: @Keats I'll leave it up to you. I left these in to match 1.x/2.x behavior, but I could see all validations going away. I don't have a strong preference for either

Keats: The only usecase for that fn is if you just want to see the data so doing the validation seems overkill. Remove it and I'll merge + release

mike-engel: Alright, makes sense. Updated now. Thanks!

Issue #57, opened 2018-07-13
Move from error-chain to failure


ufoscout:
As error-chain is currently deprecated ( https://github.com/rust-lang-nursery/failure/issues/181 ), I would propose to move to failure which is lots of traction in the rust community.

Keats:
Failure is not currently actively maintained so it's more likely this crate
will switch back to using plain Error for the next major version when a new
ring version is released.

On 13 Jul 2018 12:12, "Francesco" <notifications@github.com> wrote:

As error-chain is currently deprecated ( rust-lang-nursery/failure#181
<https://github.com/rust-lang-nursery/failure/issues/181> ), I would
propose to move to failure which is lots of traction in the rust community.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/Keats/jsonwebtoken/issues/57>, or mute the thread

ufoscout:
@Keats 

oh, sorry, for what I read I thought 'failure' was going to be the default way of handling errors in rust.
I am still a newbie in rust and I am having troubles with this simple piece of code:
```
    pub fn generate(&self) -> Result<String, failure::Error> {
        jsonwebtoken::encode(&self.header_default, &my_claims, &self.secret.as_ref())
    }
```
I use failure and I am not able to integrate the result of the jsonwebtoken::encode call with the rest of my code, the compilation fails with:
```
expected struct `failure::Error`, found struct `core::jwt::jsonwebtoken::errors::Error`
```
but I don't want to return errors that do not match the failure::Error trait, do you have any suggestion? Is there any simple way to convert from `core::jwt::jsonwebtoken::errors::Error` to `failure::Error`?

Keats:
can you do something like `.map_err(|err| err.into())` to convert errors from one to the other?

Keats:
Remove error chain in favour of plain Error

PR #58, opened 2018-07-13
Update ring

matthauck:


Keats:
It's not that simple sadly. There can only be a single version fo ring across someone project so we have to wait until a non-alpha release as otherwise every alpha release of ring would be a major version of `jsonwebtoken` as well

matthauck:
Yeah, it was a ring-asm version conflict from upgrading ring in another
project that led me to need to update it for jsonwebtoken too. Unfortunate
that there is such a tight coupling. Hopefully ring-asm doesn’t change too
frequently?

Keats:
Ring 0.13 was actually released Friday but https://github.com/briansmith/ring/pull/619 is not in it and the repo Cargo.toml is still showing alpha.
A major version will require #57 to be done as well so I'll have a look next week

Issue #21, opened 2016-10-11
Add ES* family


Keats:
ES256, ES384 and ES512

pimeys:
Hey,

Have you been investigating this? We need it for Apple push notifications and are looking into ways of implementing them in Rust. We could fork and use some C bindings to get the job done, but I guess the changes would never get into the mainline here.

There are options to do the signing with:

* https://github.com/apoelstra/rust-secp256k1/ A C binding for libsecp256k1 from the bitcoin project. At least for Arch the library is only in AUR.
* https://github.com/briansmith/ring/blob/master/src/ec/eddsa.rs Ring seems to have support for EDDSA, but I couldn't find anything for ECDSA.
* OpenSSL has some support for curves, but I'm not a crypto expert to really find the correct way of using it.

Is there any progression or people who've been trying to do this with Rust?

Keats:
I didn't look into it yet but from a very quick research, it seems that Ring does support verifying ECDS https://briansmith.org/rustdoc/ring/?search=ECDSA but not signing.
https://github.com/briansmith/ring/issues/207 seems to be tracking it

I'm not a crypto expert either so I definitely don't want to build it myself

pimeys:
I asked about ECDSA in #rust-crypto yesterday and they gave me this: https://gitlab.com/ilari_l/btls/tree/master I'll investigate how and where should I use it and at least fork your project for basis. I don't know do you want to have this include in the HEAD, but we're in serious need for this with Apple notifications.

briansmith:
ECDSA signing with P-256 and P-384 is in ring 0.13.0-alpha4. See the comments in https://github.com/briansmith/ring/issues/207 for details. Please try it out and let me know if you run into any issues before I release the 0.13.0 final version.

Keats:
Thanks for the notice!
I'll update the crate once the version is released with https://github.com/briansmith/ring/pull/619 so upgrading ring is not a breaking change

Keats:
I don't have the bandwidth to work on that but I would welcome a PR

manifest:
I'm willing to work on this issue, but there are currently some obstacles in **ring** API. I've described them in the related [issue](https://github.com/briansmith/ring/issues/207).

manifest:
@Keats Do you mind changing type of key from `&[u8]` to `enum` ?

```rust
enum Key {
    Bytes(Vec<u8>),
    ECDSAKeyPair(signature::ECDSAKeyPair),
}
```

[...]

PR #59, opened 2018-07-25
Next

Keats:


saschagrunert:
Yeah, the update of ring is needed for the current version of actix-web

Keats:
I'd like to add #21 in it as well before publishing as it is a breaking change

PR #73, opened 2019-02-01
ECDSA signing and verification (ES256, ES384)

jbg:
Also updates Ring to 0.14 and updates the RSA signing code to match Ring API changes.


jbg:
I was emailed a comment on this pull request asking why I used _ring_'s `ECDSA_P256_SHA256_FIXED_SIGNING` instead of `ECDSA_P256_SHA256_ASN1_SIGNING`, and indicating that it makes this incompatible with the keys generated by OpenSSL. However, the comment doesn't appear here. Perhaps it was deleted, but I'll answer it anyway in case GitHub has just eaten the comment somehow.

The relevant spec is [JSON Web Algorithms Section 3.4](https://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-40#section-3.4) which specifies a simple concatenation of R and then S, which is what `ECDSA_P256_SHA256_FIXED_SIGNING` gives, as indicated in the `ring` docs [here](https://briansmith.org/rustdoc/ring/signature/index.html#ecdsa__fixed-details-fixed-length-pkcs11-style-ecdsa-signatures).

So the short answer is, using `ECDSA_P256_SHA256_ASN1_SIGNING` would create invalid JWT signatures.

The question of what format the keys are provided in is pretty much orthogonal to how the signature is encoded. I chose to take them as PKCS#8 keys because [that's what `ring` accepts](https://briansmith.org/rustdoc/ring/signature/struct.EcdsaKeyPair.html) and because that's what Apple gives as APNS keys, and being able to authenticate to APNS using JWT was my primary reason for implementing this in the first place.

jbg:
It's also worth noting that OpenSSL provides tools for encoding keys in PKCS#8 structures, so it should not be a problem to use the OpenSSL tools to create suitable keys. I haven't tried, though.

briansmith:
> It's also worth noting that OpenSSL provides tools for encoding keys in PKCS#8 structures, so it should not be a problem to use the OpenSSL tools to create suitable keys. I haven't tried, though.

I documented exactly how to generate PKCS#8 formatted keys using OpenSSL that interop with *ring* at at https://gist.github.com/briansmith/2ee42439923d8e65a266994d0f70180b.

manifest:
Actually, the public key generated with instructions above isn't in a format that **ring** expect.

```bash
openssl ecparam -name prime256v1 -genkey -noout -out private_key.pem
openssl ec -in private_key.pem -pubout -out public_key.pem

openssl pkcs8 -in private_key.pem -topk8 -nocrypt -outform DER -out private_key.p8.der
openssl pkey -pubout -inform der -outform der -in private_key.p8.der -out public_key.p8.der
```



At the moment, the best solution I came with is to use **openssl asn1parse**. Another way is to use ASN1 parser to extract the public key in the application.

```bash
openssl ecparam -name prime256v1 -genkey -noout -out private_key.pem
openssl pkcs8 -in private_key.pem -topk8 -nocrypt -outform DER -out private_key.p8.der

openssl ec -in private_key.pem -pubout -out public_key.pem
openssl asn1parse -in public_key.pem -offset $((23 + 2)) -out public_key.p8.der.block
dd bs=1 skip=1 if=public_key.p8.der.block of=public_key.p8.der
```


At first, I missed that leading zero in asn1parse output, so I thought that the problem in the signing algorithm. But the example above just works with both ring's ECDSA_P256_SHA256_FIXED_SIGNING and ECDSA_P256_SHA256_ASN1_SIGNING, that is why I removed my comment in order to don't create more a confusion then there is, but probably it worth mentioning in case someone else will be confused in the similar way.

gdamjan:
is there any rust create that can convert a pem public key file to what ring expects?
I'm getting the key from [AWS dynamically](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-authenticate-users.html#user-claims-encoding)

manifest:
@gdamjan you can implement it by yoursef using **pem** and **yasna** crates.

```rust
fn parse_pem(input: &[u8]) -> Result<Vec<u8>> {
    use failure::SyncFailure;

    let pem_contents = pem::parse(input).map_err(SyncFailure::new)?.contents;
    let asn_contents = yasna::parse_der(&pem_contents, |reader| {
        reader.read_sequence(|reader| {
            reader.next().read_sequence(|reader| {
                reader.next().read_oid()?;
                reader.next().read_oid()?;
                Ok(())
            })?;
            let bytes = reader.next().read_bitvec()?;
            return Ok(bytes)
        })
    }).map_err(SyncFailure::new)?;

    Ok(asn_contents.to_bytes())
}
```

briansmith:
> @gdamjan you can implement it by yoursef using **pem** and **yasna** crates.

It looks like it's `SubjectPublicKeyInfo` format, right? If so, we can move the SPKI logic from webpki into *ring* so that this can be done without yasna (but `pem` would still be needed). This is *ring* issue https://github.com/briansmith/ring/pull/378.

manifest:
@briansmith I'm not sure about it, probably it's not

```bash
openssl asn1parse -in public_key.pem
    0:d=0  hl=2 l=  89 cons: SEQUENCE
    2:d=1  hl=2 l=  19 cons: SEQUENCE
    4:d=2  hl=2 l=   7 prim: OBJECT            :id-ecPublicKey
   13:d=2  hl=2 l=   8 prim: OBJECT            :prime256v1
23:d=1 hl=2 l= 66 prim: BIT STRING
```

manifest:
@Keats Is there something that need to be done on this one?

Keats:
Mostly the discussion on  #76

src/crypto.rs
```diff
@@ -18,6 +18,12 @@ pub enum Algorithm {
     /// HMAC using SHA-512
     HS512,
 
+    /// ECDSA using SHA-256
+    ES256,
+
+    /// ECDSA using SHA-384
+    ES384,
```
Keats: ring doesn't have ES512?

manifest: It doesn't.
https://github.com/briansmith/ring/blob/master/src/ec.rs#L37-L41

PR #74, opened 2019-02-04
Attempt to fallback to pkcs8 if der fails

Jake-Shadle:
While attempting to get https://github.com/mozilla/sccache/pull/272 working since we use GCS, I found that the reason it was failing was because jsonwebtoken was only using the `from_der` constructor for RSAKeyPair, while GCP private keys are stored in pkcs8, and falling back to `from_pkcs8` allows the key to be deserialized correctly.

I considered adding a more functions to instead allow someone to pass in the RSAKeyPair from outside the crate instead of this, but seemed like a bit of overkill when it could just be this one-liner, but let me know if you want me to rework it. :slightly_smiling_face:


Jake-Shadle:
That's odd, for reference I know for sure that [this code](https://github.com/Jake-Shadle/sccache/blob/gcs/src/cache/gcs.rs#L227-L237)

```rs
        // Could also use the pem crate, but that seems overly complicated for just the specific
        // case of GCP keys
        let key_string = self.sa_key.private_key.splitn(5, "-----").nth(2).ok_or_else(|| "invalid key format")?;
        // Skip the leading `\n`
        let key_bytes = base64::decode_config(key_string[1..].as_bytes(), base64::MIME)?;

        let auth_request_jwt = jwt::encode(
            &jwt::Header::new(jwt::Algorithm::RS256),
            &jwt_claims,
            &key_bytes,
        )?;
```

works correctly when used against the code in this PR, as I go from 100% cache misses due to being unable to acquire a token to it functioning normally, but without the previous dependency on OpenSSL (just to do the conversion to DER).

Keats:
Looks like `signature::RSAKeyPair` is not at the same place anymore in ring 0.14?

Keats:
Closed in favour of #89

PR #78, opened 2019-03-09
Use ring 0.14

dvic:
I ran into problems because `rustls` relies on `ring 0.14` and `jsonwebtoken` relies on `ring 0.13`. I'm not sure what the plan is for moving to `ring 0.14` but here is a PR for it.


dvic:
🤦‍♂️ did not see #75, never mind 😆

Issue #76, opened 2019-02-04
v7 discussion


Keats:
There are quite a few changes happening in the PRs: more input format for RSA (#69, #74), ECDSA signing & verification (#73) as well as some Validation changes.

I have already removed the `iat` check in the `next` branch since it isn't something that should be checked.

Right now, `Validation::algorithms` is a `vec`. I don't remember why but it shouldn't be the case, it should be an `algorithm: Algorithm` instead, I will accept a PR for that or do it myself later.

#69 also adds a standalone `verify_rsa` fn, which I'm wondering if we can somehow streamline it with the rest of the crate.

Since Rust doesn't have default parameters, we always have to pass a `Validation` struct currently. Maybe we can put the decoding onto this struct instead so you get something like:

```rust
// currently
let token = decode::<Claims>(&token, "secret".as_ref(), &Validation::default())?;

// possible
// `JwtDecoder` has the same fields as the current `Validation`
let token = JwtDecoder::default().decode_hs256::<Claims>(&token, "secret".as_ref())?;
```
This way we don't have a single function where we are trying to fit all arguments inside and the user has to select explicitely which decode fn to use. This solves the `vec` of algorithms at the same time and allows having better documentation for each. The downside is duplication of those functions/docs for the various digest values (`decode_hs256`, `decode_hs384`, `decode_hs512` for each algo).

Any other ideas/criticisms/things missing?

ccing the people involved in various issues/PRs recently
@AaronFriel @jbg @Jake-Shadle @matthew-nichols-westtel @greglearns

Jake-Shadle:
I only became aware of this lib yesterday because of the bug I was tracking down, and while #74 works, it's not what I would call clean.

So my one thought looking at this code only briefly was that it might be nice to be able to expose some helpers for deserializing the RSA key based on the user knowledge and passing that key down to the API rather than the current method of just trying both. The reason I say helpers would be nice instead of just passing the key directly is that would require the user adding 2 new dependencies, `ring` and `untrusted`, though another option would of course to be to expose the types and functions needed from those crates as part of the public API of this lib.

briansmith:
[...]

PR #79, opened 2019-03-15
chore: upgrade ring and base64

Brooooooklyn:
Please bump a patch version if you merged this pr.
A temporary here:
```toml
[replace]
"jsonwebtoken:5.0.1" = { git = "https://github.com/Brooooooklyn/jsonwebtoken.git" }
```


Keats:
It would actually need to be a major version bump if you bump ring.
It is already done in #75 and blocked a bit on #76

PR #83, opened 2019-04-01
Add example of using `chrono::DateTime` in claims

### LINKED ISSUE #82: Provide example using chrono::DateTime
`chrono` is a very popular date and time library and defaults to serializing as ISO strings instead of timestamps as per the JWT spec. It'd be nice to have an example using the `#[serde(with = "..")]` attribute. I can submit a PR with this example if it is accepted for inclusion.

rakenodiax:
Using `chrono`'s `serde` feature uses ISO 8601 instead of a Unix 
timestamp as specified in RFC 7519 section 2, "NumericDate". This 
example uses  custom de/serialize functions as shown in the [serde.rs 
example, "Custom Date Format"](https://serde.rs/custom-date-format.html).

NOTE: Currently fractional values are not supported in the example, 
though they are in the spec.

Resolves #82

### DIFF
--- a/examples/custom_chrono.rs
+++ b/examples/custom_chrono.rs
@@ -1,0 +1,101 @@
+extern crate jsonwebtoken as jwt;
+extern crate serde;
+#[macro_use]
+extern crate serde_derive;
+extern crate chrono;
+
+use chrono::prelude::*;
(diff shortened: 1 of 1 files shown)


Keats:
Late to get to that, sorry!

PR #86, opened 2019-04-20
Update ring dependency to 0.14.6, refactor sign_rsa fn

azah:
I ran into a dependency issue because actix-http requires ring 0.14+. I bumped the version and adjusted the rsa function as needed.

All tests pass, I basically just changed it to the new example in ring's documentation.

* Ring 0.14.6 no longer has a signing state. The
  key_pair itself can call #sign.
* Ring 0.14.6 also doesn't have the rsa feature
  anymore.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ keywords = ["jwt", "web", "api", "token", "json"]
 serde_json = "1.0"
 serde_derive = "1.0"
 serde = "1.0"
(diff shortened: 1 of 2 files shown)


Keats:
Is there a difference with #75

saks:
Can this PR be merged in order to get ring 0.14 out as a dependency quicker? Any blockers here?

azah:
@Keats the only difference is the version: 14.6 (PR) vs 14.4 (master, now that you merged it).

Either works though, thanks!

PR #87, opened 2019-04-30
Jwt

### LINKED ISSUE #77: Decoding with x509 certs
I'm having a hard time authenticating a token using a x5c.  (MS OAuth/Azure) 

Below is the code... 

```rust
// Trying to isolate the problem by only checking the signature. 
let validation_config = jsonwebtoken::Validation {
            algorithms: vec![jsonwebtoken::Algorithm::RS256],
            leeway: 0,
            validate_exp: false,
            validate_iat: false,
            validate_nbf: false,
            aud: None,
            iss: None,
            sub: None
        };
let token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsIn...";
let x5c_cert = "MIIDBTCCAe2gAwIBAgIQKOfEJNDyDplBSXKYcM..."; 

let raw_der = base64::decode_config(der, base64::STANDARD).unwrap();
let d = jsonwebtoken::decode::<MsOAuthPayload>(&token, &raw_der, &validation_config);
```

The above always returns InvalidSignature.  

* RS265 is the correct algo. 
* The cert is correct. I tried it on jwt.io by adding a BEGIN/END cert and it validates fine. 
* I used ssl to convert the BEGIN/END pem to DER and the bytes match up from the base 64 decode. 
* The key URL is: https://login.microsoftonline.com/common/discovery/v2.0/keys but my specific tenant returns the same keys. 

Anyone have some insight on what I'm doing wrong here? 

Thanks

### LINKED ISSUE #85: Keys in test folder
Please specify that in the test folder, the correct public key for `private_rsa_key.pem` is `public_rsa_key_8.pem` and not `public_rsa_key.pem`. Looking at the tests where the `*.der` files are one would assume that it's `public_rsa_key.pem` that is the correct one but I only got the test token verified by using `public_rsa_key_8.pem` and thought there was something wrong in my `base64` to bytearray conversion :)

gpit2286:
Closes #77 
Closes #85


Keats:
The next version is changing things considerably so closing this, it might be pulled later though.

PR #88, opened 2019-05-10
Fix algorithm mapping.

maxburke:
RS256/384/512 algorithm strings were mapped to HMAC-SHA256/384/512
enumerators.


Keats:
Whoa thanks for noticing and fixing that!

PR #89, opened 2019-05-15
Clarify Key formats

Jake-Shadle:
This is a bit of a rethink of my old PR #74 that I think improves the API to be more clear for users by forcing them to specify the key format (this is fudged a bit for HMAC) that they are using. This makes the API more explicit and is a breaking change, but it's pretty trivial to update calls so I think it's an overall win, but I'm open to suggestions for making it more ergnomic.


Keats:
I like that a lot! Much cleaner than my initial idea.
I think it needs to add key components as well (https://github.com/Keats/jsonwebtoken/pull/69) but that can be done in another PR.
I'll start another branch for breaking changes and merge that in asap

PR #93, opened 2019-06-07
Support for RSASSA-PSS signing algorithm

### LINKED ISSUE #92: Support for RSASSA-PSS signing algorithm
As specified in https://tools.ietf.org/html/rfc7518#section-3.5

-  PS256 - RSASSA-PSS using SHA-256 hash algorithm
-  PS384 - RSASSA-PSS using SHA-384 hash algorithm
-  PS512 - RSASSA-PSS using SHA-512 hash algorithm

The RSASSA-PSS signature verification algorithm is supported by ring

kanru:
As specified in https://tools.ietf.org/html/rfc7518#section-3.5

- PS256 - RSASSA-PSS using SHA-256 hash algorithm
- PS384 - RSASSA-PSS using SHA-384 hash algorithm
- PS512 - RSASSA-PSS using SHA-512 hash algorithm

Fixes #92 

This is a breaking change so it's a good chance to release with v7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,9 @@ This library currently supports the following:
 - RS256
 - RS384
 - RS512
+- PS256
+- PS384
+- PS512
 - ES256
 - ES384
 
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -30,6 +30,13 @@ pub enum Algorithm {
     RS384,
     /// RSASSA-PKCS1-v1_5 using SHA-512
     RS512,
+
+    /// RSASSA-PSS using SHA-256
+    PS256,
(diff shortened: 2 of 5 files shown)


Keats:
Thanks! What format do they keys for that algorithm are like? der/pem or do you also get modulus/exponent sometimes?

kanru:
@Keats the algorithm uses same key format as other RSA algorithms. Ring currently can only use keys with rsaEncryption algorithm ID.

I would like to use keys in pkcs#8 format. Looks like it will be supported by v7.

Keats:
Thanks!

PR #95, opened 2019-06-25
Add a NONE signature algorithm.

otaconix:
This algorithm results in empty signatures when signing. Verification
always yields success, whether the signature is empty or not.

PS: apologies for the unrelated formatting changes. rustfmt ran automatically, and this is what it spat out :)


Keats:
The None algorithm is intentionally not implemented since there are no reasons to use it. It's equivalent to https://docs.rs/jsonwebtoken/6.0.1/jsonwebtoken/fn.dangerous_unsafe_decode.html minus the validation but if you don't have a signature, there is nothing to validate....

PR #97, opened 2019-08-11
Update Ring to 0.16

afldcr:
I need some features out of the newer version of Ring, and everything seems to work great after updating and renaming a few things. I know the PR for v7 is up, but this may hold people over until it's finished.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "jsonwebtoken"
-version = "6.0.1"
+version = "6.0.2"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "jsonwebtoken"
-version = "6.0.1"
+version = "6.0.2"
```
Keats: that's the issue, updating ring is a breaking change so that would have to be v7

ErichDonGubler:
Since this isn't going to actually get merged, would it be fair to close it? People can use the `next` branch or a commit therein for now if they want these changes.

PR #100, opened 2019-09-27
Update ring to 0.16

mwilliammyers:
- Update `ring` to 0.16
- Add `std` feature to `ring` (so errors impl `std::error::Error`)
- Remove the now unused `untrusted` dependency, as `ring` now handles that internally.


mwilliammyers:
Just realized this was done as a part of #91. I will leave this open in case you want to bump ring separately from the rest of #91.

PR #105, opened 2019-10-27
Updated aud within Validation, including membership validation testing

Dowwie:
- Validation.aud is now `Option<HashSet<String>>`
- audience validation tests for "any of these" membership

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Add support for verifying with modulus/exponent components for RSA
 - Change API for both sign/verify to take a `Key` enum rather than bytes
 - Update to 2018 edition
+- Changed aud field type in Validation to Option<HashSet<String>>.  Audience 
+  validation now tests for "any-of-these" audience membership.
 
 ## 6.0.1 (2019-05-10)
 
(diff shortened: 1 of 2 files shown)


Keats:
What I was trying to avoid with `add_audience` is to not have `Value` be in the struct at all. It was there only since it could be a list or a string, now that this is a HashSet, there is no reason to still have it. The name can stay `set_audience` to avoid thinking about incrementing something.

Eg you can have

```rust
validation.set_audience(&"my_corp");
// OR if you have several audiences
validation.set_audiences(&["corp1", "corp2"]).
```

with types
```rust
set_audience(name: &str)  // or impl ToStr, whatever

set_audiences(names: &[&str]) // or impl ToStr, whatever
```

This removes all Value from the interface, the `unwrap/panic` as well as the `convert_aud` in your PR.

Dowwie:
@Keats I'm under the impression that this PR is ready for approval

Keats:
The CI failures are legit, you removed the `crate::` from the imports and v7 is in 2018 edition so those are needed.

Dowwie:
@Keats   oh ok.  fixed.

Keats:
Thanks!

PR #106, opened 2019-10-30
Add PEM decoding support

LeviSchuck:
This adds a new top level function `decode_pem`, and a subsequent function `as_key` on the result type for use with `sign` and `verify`.

This contributes to #76 and #77 
However it does not include X.509 certificate support.

I have an idea on how to do that but this should handle many cases that users want as PEM files are the common output from `openssl`

It supports both PKCS#1 and PKCS#8 RSA keys, it supports PKCS#8 EC keys, however it doesn't assert that the curve ID but within JWT contexts the curve is predefined by the spec.

PR #107, opened 2019-11-03
Add PEM decoding support (#106)

Keats:
* Add PEM support with pem and simple_asn1. Documentation TODO

* Make pkcs1 and pkcs8 versions of the RSA key, confirm they pass tests.

* Add documentation, simplify

* Update readme

* Bump pem version

* Remove extra print


Keats:
@LeviSchuck 

b9a3e30  removes the `Option::`, it's in the prelude so no need for them
06bebea  runs `cargo fmt`, definitely something very nice to use, the original code was 2 spaces indented rather than the normal 4
caef740 does some refactoring to avoid some duplication as well as changing the names of enum members: even if they are abbreviations like RSA, it is more idiomatic to have them as `Rsa`

I'll start thinking about the new API now, with pem handling it can be much nicer than the current `next` branch.

LeviSchuck:
I've reviewed the 3 commits you've referenced and I'm good with them.

Keats:
So my thought for the api would be to take a `&str` again for both hmac and pem. The main issue is the Rsa(n, e) type of public key, I wonder if we can easily build a pem string from modulus/exponent with ring.

LeviSchuck:
It is possible. 
If you drop the public key ( https://github.com/Keats/jsonwebtoken/blob/pem-handling/tests/rsa/public_rsa_key_pkcs1.pem ) into https://lapo.it/asn1js/

image

You'll see that it's encoded as a ASN1 sequence of a big endian big int followed by the modulus (similarly encoded)

The simple_asn1 library already included may be able to assist with that.

LeviSchuck:
In fact, it's one of the things I was looking into, using another JWK to feed these algorithms which rely on decoding an ASN1 body. 

The challenge I had with the simple asn1 though is that it wants the user to calculate the size of things.
https://docs.rs/simple_asn1/0.4.0/simple_asn1/enum.ASN1Block.html

I haven't gotten to the part of actually doing it, but in theory it would be `to_der(ASN1Block::Sequence(?,vec!(ASN1Block:: Integer(?,modulus), ASN1Block:: Integer(?,exponent))))`

Keats:
Nice, so I guess the encode/decode with `&str` is doable, just need to have a function that takes exponent/modulus and returns a pem and users can pass that to encode/decode.

Keats:
@LeviSchuck are you interested in building a function that looks like `(exponent: &[u8], modulus: &[u8]) -> Result<String>` in `pem_decoder.rs` (for now). Otherwise I'll add it when I have time but you know more about asn1 than I do. I had a quick look:

```
pub fn ne_to_pem(modulus: &[u8], exponent: &[u8]) -> Result<String> {
    let der = to_der(&ASN1Block::Sequence(
        0,
        vec![
            ASN1Block::Integer(0, BigInt::from_signed_bytes_be(modulus)),
            ASN1Block::Integer(0, BigInt::from_signed_bytes_be(exponent)),
        ],
    ))
    .unwrap();
    let encoded = base64::encode(&der);
    let formatted = format!("-----BEGIN RSA PUBLIC KEY-----\n{}\n-----END RSA PUBLIC KEY-----", encoded);

    Ok(formatted)
}
```
This generated a valid key (afaik) but I could not get the test `rsa_modulus_exponent` test to pass, maybe because of the offset or me not having read enough about asn1/der.

LeviSchuck:
Hey @Keats 

Thanks for the example code. I've spent a few weekends at a coffee shop studying this stuff for that exact reason. I didn't use simple_asn1 for it, its much more restricted and hacky in my current implementation but I can put some time tonight into this.

For reference I was making a way to go to and back from pem and jwks

```

Keats:
No hurry, take your time. If we nail that function, we can have a pretty optimal DX while supporting more formats than before.

LeviSchuck:
It would also mean that I can drop my inflexible jwt implementation :)

LeviSchuck:
The API could definitely use some refining or access via module, but I got it working for public keys. (EC(x) and RSA(n,e))

https://github.com/LeviSchuck/jsonwebtoken/commit/311192a2431a393ff8d70a5a19f7ec45e81b5d2f

Everything is in one commit so far but I'd make a PR once I get documentation in. Then @Keats you can make the API more DX friendly.

Keats:
That looks very good, thanks for that. Some changes will be required (eg not exposing `ring`) but can be done later. Is anyone actually using EC JWK? I don't think I've seen any of that myself.
Don't worry about the documentation, it will probably change shape a lot so the documentation might be outdated.

LeviSchuck:
I haven’t seen it yet, but I’m wanting to use it for my own projects.
The specification says that EC is optional and RSA is expected, but also TLS 1.3 dropped RSA in the cipher suites supported.

As for not exposing ring, that’s only in the tests as a convenience to prove that I can go from private PEM to ring to public pem, else I’d need to start making “key parameters” and that’s kinda reimplementing ring. I could also hard code the keys in the tests but that’s less confidence to me than using arbitrary external key data via include str

DX wise I think the next part is going from modulus exponent to der for rsa

LeviSchuck:
I also replied on the commit, the specification predefined the curve used for JOSE HS* as prime256v1

Keats:
> As for not exposing ring, that’s only in the tests as a convenience to prove that I can go from private PEM to ring to public pem, else I’d need to start making “key parameters” and that’s kinda reimplementing ring

Those tests are fine but should be implemented in the module itself.

> DX wise I think the next part is going from modulus exponent to der for rsa

I am thinking into having dedicated decode functions for RSA jwk and EC jwk instead of using the default `decode` so the users just put n/e or x in some struct/tuple and everything happens internally (including jwk sets I guess?).

If you can do a PR with what you currently have on this branch, we can start having a look.

LeviSchuck:
I noticed while peeking into ring, there were functions that took the parameters directly.

But sure, I’ll submit a pr

LeviSchuck:
👍

Keats:
Important tools for Rust: clippy and fmt as shown in a6ea8c2

Keats:
I have the first batch of change with the commit above (e3632b3a2c874b6112516d23b4a952a2305a570f) which has encode/decode now only taking a hmac key or pem bytes. No more `der` support so likely need to expose one of the pem_encode function to ease the transition from previous versions.
I'll handle JWK/JWK sets next but don't hesitate to comment already.

LeviSchuck:
One note that I hit when doing my changes earlier,

The RSA Modulus from ring doesn't have a leading zero (and the function to get it specifically is named such), but in a PEM it does have a leading zero. I'm not sure if ring itself will take one without a leading zero.

LeviSchuck:
RE: JWKs are you planning on doing verification with a JWK struct or vec of jwks?

Keats:
After some thoughts, I might just have an example for JWK set since they can in people own struct/hashmap etc and it is probably simpler to provider the plumbing + example on how to do it

Keats:
I've added not yet working JWK decoding, I'm not entirely sure why https://github.com/Keats/jsonwebtoken/pull/107/commits/b27981549f93a0f667e064cd3c7e2956a6a97277#diff-d866210ac5208ba9071685e86eba4beeR117-R127 is failing to verify but that can be left for later.

LeviSchuck:
Its probably what I mentioned earlier

> The RSA Modulus from ring doesn't have a leading zero ... but in a PEM it does have a leading zero. 

I unfortunately don't have much time to poke at this.

Keats:
Look at https://github.com/Keats/jsonwebtoken/pull/107/commits/614f3610a72b752d9241b06bcaf7c1f27910400a to see the bugfix...

🤦‍♂

LeviSchuck:
Oh, I can see how that went wrong. Glad that's resolved.

Keats:
@LeviSchuck after writing the docs, I've realized I don't actually need anything from `pem_decoder.rs` since there is a method that takes `(n, e)` specifically.
Do you have a usecase for it in your own code or can I remove it for now?

Keats:
I think API & docs is in a good place right now, probably ready for some alpha release after I add a few more testcases from jwt.io.

Keats:
Thanks a lot @LeviSchuck !

I'll push an alpha release now for v7

rokit:
Hi, I'm extremely new to all this, and I'm following this guide for verifying a google token.
https://ncona.com/2015/02/consuming-a-google-id-token-from-a-server/

I'm stuck here:
> This key (the `kid` key in the header) can’t be used directly to verify the token signature. First we need to transform it to a PEM key. I’m not going to explain how to do this because there is already a node module that can do it for us called `rsa-pem-from-mod-exp`.

Does this library allow verifying with PEM and/or transforming `kid` to PEM with a modulus and exponent?

LeviSchuck:
Hey @rokit 

You'll need to fetch the remote JWK, find the key associated to the kid present in the JOSE header, and then use the parameters in that JWK to create a public key. You can then verify the key from there. 

A constructor for a public key of that kind should be demonstrated in the test folder.

rokit:
Thanks @LeviSchuck. It looks like I needed `decode_rsa_components`.

PR #108, opened 2019-11-06
Add functions to create pems and ders of the public keys

LeviSchuck:
It’s dirty and lacks documentation, tests probably can be relocated, but now a user of this library can make DER and PEMs of public key parameters


Keats:
Cheers!

PR #113, opened 2019-12-29
Add EncodingKey & DecodingKey

Keats:


Dowwie:
is there a pressing reason to use borrows?  the lifetime requirements cascade all over my code base

Keats:
For EncodingKey not really, we could remove the lifetime. For DecodingKey we do not want to clone the secret/keys everytime we verify a token so they are needed.

Dowwie:
@Keats The DecodingKey isn't using a secret but rather a public key.  Clone away, if the concern is security.

DecodingKey is the bigger problem for me as I use it in every system.

Keats:
Why not instantiate it in a lazy_static or similar? The whole point of that API is to allow re-using the Encoding/Decoding keys to make it more performant.

Dowwie:
I'm willing to compromise with a lazy_static but haven't figured out an impl to satisfy the compiler without making some slight changes to DecodingKey::from_rsa_pem, changing the key param to String.  Without changing this to String, I couldn't get lazy_static to work (temporary value exceptions).
```
  /// If you are loading a public RSA key in a PEM format, use this.                                                                                    
 53     pub fn from_rsa_pem(key: String) -> Result<Self> {                                                                                                    
 54         let pem_key = PemEncodedKey::new(key.as_bytes())?;                                                                                                
 55         let content = pem_key.as_rsa_key()?;                                                                                                              
 56         Ok(DecodingKey { kind: DecodingKeyKind::SecretOrDer(Cow::Owned(content.to_vec())) })                                                              
 57     }   
```
an example of the lazy_static:
```rust
lazy_static! {
    static ref JWT_PUBLIC_KEY: DecodingKey<'static> = 
       DecodingKey::from_rsa_pem(
          fs::read_to_string(
            env::var("JWT_PUBKEY")
            .unwrap_or_else(|_| panic!("Failed to load JWT public key envvar"))
          ).unwrap()
       ).unwrap_or_else(|_| panic!("Failed to create JWT DecodingKey"));
}
```

Dowwie:
What do you think about modifying the from_rsa_pem functions accordingly?

Keats:
I would prefer keeping `&[u8]` for maximal flexibility. 
In your lazy_static example, I think you need to define another variable for the env var:

```rust
lazy_static! {
    static ref SOME_STRING: String = env::var("JWT_PUBKEY").unwrap();
    // And there you can have your DecodingKey with &[u8] I think
    static ref OTHER: &'static [u8] = SOME_STRING.as_bytes();
}

```

Dowwie:
correct!   thanks for the lifetime tip and for simplifying EncodingKey.  I'll get by with &[u8].  I'm running tests using the latest changes through my systems.  If you don't hear back from me shortly then that means no further concerns.

```rust
lazy_static! {
    // borrows within DecodingKey are used for performance reasons, so this 
    // static is necessary in order to avoid lifetime baggage across code base
    static ref _PUBKEY: String = {
       let path = env::var("URBANIS_JWT_PUBKEY")
            .unwrap_or_else(|_| panic!("Failed to load JWT public key envvar"));
       fs::read_to_string(path)
          .unwrap_or_else(|_| panic!("Failed to read JWT public key to string"))
    };

    static ref JWT_PUBLIC_KEY: DecodingKey<'static> =  {
       DecodingKey::from_rsa_pem(_PUBKEY.as_bytes())
            .unwrap_or_else(|_| panic!("Failed to create JWT DecodingKey"))
    };
}
```

Dowwie:
@Keats I think I'm creating PEM files in an unsupported format.  what command are you using to generate your PEM files?  I was using this this format is failing:
`openssl req -x509 -days 365 -nodes -newkey rsa:2048 -keyout privkey.pem -out cert.pem `

Keats:
x509 is not supported

Something like:

```
openssl genrsa -out private.pem 2048
openssl rsa -in private.pem -outform PEM -pubout -out public.pem
```

Dowwie:
I think this API will work for me.  No other issues to report, yet.  My tests are passing.

Dowwie:
@Keats this PR seems sufficient for my needs.. no further issues/concerns to report

Dowwie:
@Keats essentially, what are we achieving with these changes?

Keats:
You ensure the key is of the same type as the algorithm so you can't misuse it. Eg you want to decode a RSA token but put a HMAC key instead and it will fail immediately. It's not prevented at compile time sadly but that's better than nothing. I don't think we can have type safety at compile time without have one encode/decode function per algorithm family

PR #116, opened 2020-01-29
Use newest version of ring

MarcAntoine-Arnaud:
using version 0.16.9 of ring fixes the build for WebAssembly target of JsonWebToken.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2018"
 [dependencies]
 serde_json = "1.0"
 serde = {version = "1.0", features = ["derive"] }
-ring = { version = "0.16.5", features = ["std"] }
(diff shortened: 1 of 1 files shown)


Keats:
No need to change that, you should already get the latest 0.16 version if you cargo update

MarcAntoine-Arnaud:
the version is hardcoded `"0.16.5"` to be able to use newest version you have to change with:
`"^0.16.5"`.

Keats:
Cargo default to `^` for resolution so with `ring = "0.16.5", it will take the latest version of `0.16.x` which is currently `0.16.9`. You can check by looking at the Cargo.lock generated.

MarcAntoine-Arnaud:
Hoo okay I have fixed it using the latest version (7.0.0) released recently.
It works now !
Thank you

PR #117, opened 2020-01-31
Fix recursive Display::fmt implementation on Error to avoid stack overflow

fujiapple852:
Invoking `Error::to_string()` results in a stack overflow for various error types.  This is a simply fix to avoid the recursive call in `Display::fmt`.


Keats:
Thanks!

PR #119, opened 2020-02-03
Add jwk header field

keorn:
The `Header` struct was missing on of the fields defined in the RFC: https://tools.ietf.org/html/rfc7515#section-4.1.3

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -27,6 +27,11 @@ pub struct Header {
     /// Defined in [RFC7515#4.1.2](https://tools.ietf.org/html/rfc7515#section-4.1.2).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub jku: Option<String>,
+    /// JSON Key
+    ///
+    /// Defined in [RFC7515#4.1.3](https://tools.ietf.org/html/rfc7515#section-4.1.3).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub jwk: Option<String>,
     /// Key ID
     ///
     /// Defined in [RFC7515#4.1.4](https://tools.ietf.org/html/rfc7515#section-4.1.4).
@@ -52,6 +57,7 @@ impl Header {
             alg: algorithm,
             cty: None,
             jku: None,
+            jwk: None,
             kid: None,
             x5u: None,
             x5t: None,


Keats:
Odd that I missed this for so many versions. It is theorically a breaking change but I think it's very unlikely it would break anyone? Should we add all the JWS header fields?

LeviSchuck:
> The "jwk" (JSON Web Key) Header Parameter is the public key that
   corresponds to the key used to digitally sign the JWS.  This key is
   represented as a JSON Web Key [JWK].  Use of this Header Parameter is
   OPTIONAL.

Related documentation:
https://developer.cisco.com/codeexchange/github/repo/cisco/node-jose/

> The key can be embedded using either 'jwk' or 'x5c', and can be located in either the JWS Unprotected Header or JWS Protected Header.
>
> NOTE: verify() will use the embedded key (if found and permitted) instead of any other key.

My note: *be cautious about verifying embedded keys instead of referenced keys.* Alone it doesn't provide authentication that the key is from a trusted source. jsonwebtoken already has `jku`, a url to the the jwk set, the url can be used to establish trust.

I think being able to serialize with `jwk` is trivially safe, but verifying is not safe by default without authenticating the public key. Some consideration should be given to that.

LeviSchuck:
For more about my security concern, 
* review the "Embedding Keys" section on https://www.pingidentity.com/fr/company/blog/posts/2019/the-hard-parts-of-jwt-security-nobody-talks-about.html
*  in regards to the cisco jose library: https://nvd.nist.gov/vuln/detail/CVE-2018-0114
* https://www.owasp.org/images/0/07/20190222--Nyffenegger-JWAT.pdf

Also, **`string` is the wrong type**.. It should be a an optional single JWK key. 

An example of it can be seen in https://github.com/zi0Black/POC-CVE-2018-0114/blob/master/jwk-node-jose.py (a proof of concept for the vulnerability I am cautioning about)

If this is to be added, and verification is included, I think that verification should ignore it by default and can only be enabled by an additional option which is by default **not allowing embedded JWK use**. 

Note that this PR does not include verification, but it is quite incomplete if verification is not included.

keorn:
@LeviSchuck This PR makes this library capable of ingesting compliant tokens.
None of the other `Header` fields are validated and this library is `String`ly typed in a number of other ways.

Feel free to close if you prefer not to add this field as is.

LeviSchuck:
Reviewing the types in the header prior to this PR
- `typ` always `"JWT"`
- `alg` is typed to Algorithm enum, like `"HS256"`
- `cty` are mime types, the values are strings https://tools.ietf.org/html/rfc7515#section-4.1.10
- `jku` is a url, so the value is a string
- `kid` is an opaque string to identify the key used to the audience
- `x5u` is a url, so the value is a string
- `x5t` is a base64 encoded sha1 of the X.509 cert, so the value is a string.

It appears all fields present before this PR are intended to be strings and are not `string`ly typed.

Keats:
I am guessing the issue is more when trying to decode some existing JWT the values are not deserialized so someone needs to manually b64 decode and load it if they want to access those values?

LeviSchuck:
I'm not sure what you're saying Keats. In the limited examples I am finding for an embedded jwt, it's like
```json
{
  "alg": "RSA256",
  "jwk": {
    "kty": "RSA",
    "use": "SIG",
    "kid": "opaque id here",
    "e": "b64 bignum, usually AQAB = 65537",
    "n": "b64 bignum"
   }
}
```

I've never seen 
```json
{
  "alg": "RSA256",
  "jwk": "ewogICAgImt0eSI6ICJSU0EiLAogICAgInVzZSI6ICJTSUciLAogICAgImtpZCI6ICJvcGFxdWUgaWQgaGVyZSIsCiAgICAiZSI6ICJiNjQgYmlnbnVtLCB1c3VhbGx5IEFRQUIgPSA2NTUzNyIsCiAgICAibiI6ICJiNjQgYmlnbnVtIgp9"
}
```

Keats:
I meant that the JWT itself is going to be like `b64encoded.b64encoded.b64encoded`. If the field is not in header struct, it will not be serialized and someone want to get that value would need to the following in pseudo-code: `json::loads(b64_decode(jwt.split(".")[0]))`.
I had never seen the `jwk` field in use though, if it's a dict then the PR is not right.

LeviSchuck:
Oh yes, it'd be in the first b64 segment, the header. What you are describing is the compact serialization in [JWS section 3.1](https://tools.ietf.org/html/rfc7515#section-3.1)

> BASE64URL(UTF8(JWS Protected Header)) || '.' ||
> BASE64URL(JWS Payload) || '.' ||
> BASE64URL(JWS Signature)

In the specification for [JWS Protected Header](https://tools.ietf.org/html/rfc7515#section-4), the ["jwk" field](https://tools.ietf.org/html/rfc7515#section-4.1.3) is to be a single [JWK](https://tools.ietf.org/html/rfc7517). The JWK is dict-like, having standard fields for information about the key (like X.509), leaving key specific fields up to the [algorithm parameters](https://tools.ietf.org/html/rfc7518#section-6) as additional fields. All those additional parameters can be found in the [JSON Web Key Parameters Registration](https://tools.ietf.org/html/rfc7518#section-7.5)

keorn:
For what its worth some other JWT popular libraries like [node-jsonwebtoken](https://github.com/auth0/node-jsonwebtoken) accept only strings as header field values, which means that one has to stringify JWK for it to work.

LeviSchuck:
@keorn, are you integrating with a server that is using node-jsonwebtoken and expects strings in the JWS protected header under the key `"jwk"`?

I looked at [node-jsonwebtoken](https://github.com/auth0/node-jsonwebtoken), as well as it's dependency [node-jws](https://github.com/brianloveswords/node-jws) and am not seeing an example of this, or intended limitation.

keorn:
I use `node-jsonwebtoken` in the browser to generate a JWT. I would like to include a JWK in the header so that a Rust server can read it.
When using `node-jsonwebtoken`, the library currently only allows me to include a string under `jwk` field, so this is what I would like the decode using this library. Unless we change that library (and probably some other ones in conjunction) it may not be possible to decode headers with `jwk` field with this library.

I understand that it would probably make more sense if everyone would use an object as a value.

LeviSchuck:
I think I have an alternative solution for you @keorn 
Per RFC7800 https://tools.ietf.org/html/rfc7800#section-3.2
You can use a "cnf" claim with an embedded "jwk"

Copied below is from the RFC
```json
{
  "iss": "https://server.example.com",
  "aud": "https://client.example.org",
  "exp": 1361398824,
  "cnf":{
    "jwk":{
      "kty": "EC",
      "use": "sig",
      "crv": "P-256",
      "x": "18wHLeIgW9wVN6VD1Txgpqy2LszYkMf6J8njVAibvhM",
      "y": "-V4dS4UaLMgP_4fY4j8ir7cl1TXlFdAgcx55o7TkcSA"
    }
  }
}
```

As is, I think adding a string JSON encoded JWK breaks the standard, so I do not support it there.
This will however require you to decode the jwt to extract the `cnf.jwk` claim before verifying it. 

As I caution in previous comments, embedded JWKs should be authenticated to a whitelist before being used to verify them. If you're doing dynamic user registration with a client side JWK, that's pretty neat.

keorn:
I do not want to pollute the payload with it, since I want different ways of attaching a key to payloads in the same format.

Anyway its currently somewhat useless to attach a JWK since most JWT libraries (including this one) work only with PEM format public key and there are no utilities in Rust for conversion between JWK and PEM.

Keen to see this field supported once you have a good type for it (and maybe even decoding JWTs using a JWK).

jnicholls:
This weekend I can take a stab at a well-structured Jwk type and add the header accordingly.

Keats:
@jnicholls is that still planned?
We can probably re-use most/all of https://docs.rs/biscuit/0.5.0/src/biscuit/jwk.rs.html#1-1484 and that would be amazing to have it - I think it's the only thing holding up v8.
It would be very useful to have proper JWK types so we can have a good Auth0 example for example.

Keats:
Closing in favour of https://github.com/Keats/jsonwebtoken/pull/195 please have a look and give feedback if have some!

PR #121, opened 2020-02-21
Update readme

jareddlc:
Updates the readme to include the fields that validation checks in the claims struct and lists the default values for the validation struct.

Being new to JWT, I was having issues with decoding the token, only to realize that a certain field in the Claims struct is mandatory by default. I hope this would help people in my same position.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,20 @@ struct Claims {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Keats:
Thanks!

Issue #120, opened 2020-02-21
DecodingKey is difficult to re-use  


tjwilson90:
```
impl<'a> DecodingKey<'a> {
    pub fn from_rsa_pem(key: &'a [u8]) -> Result<Self> {
        let pem_key = PemEncodedKey::new(key)?;
        let content = pem_key.as_rsa_key()?;
        Ok(DecodingKey {
            family: AlgorithmFamily::Rsa,
            kind: DecodingKeyKind::SecretOrDer(Cow::Owned(content.to_vec())),
        })
    }
}
```
This constructs a DecodingKey that owns all of its state, but the signature suggests that it borrows the input key. As a result, the only way to construct a DecodingKey<'static> for lazy_static or once_cell is to leak the input key (or unsafely transmute the lifetime of the input key and hope that DecodingKey never changes its implementation to actually.borrow the input).

Keats:
The intent was indeed to use it in a lazy_static way, something like (pseudo-code)

```
lazy_static! {
  let decoding_key = {
      let bytes = include_bytes!("public.pem");
      DecodingKey::from_rsa_pem(bytes).unwrap()
  }
}
```

What would you change?

tjwilson90:
That works, but only if I embed my public key in my program (which would require me to recompile and redeploy if I change my key, and would make it impossible for a user of my pre-compiled program to generate and use their own keys).

Since the resulting DecodingKey doesn't need to maintain a borrow on the key, the key should have a distinct lifetime, i.e., the type of key should be `&[u8]`, not `&'a [u8]`.

MayaWolf:
Just ran into this as well. It'd be way more versatile if DecodingKey's methods just borrowed or consumed the key, depending on internal implementation, rather than requiring an external guarantee about the lifetime.

Keats:
I've pushed a stop-gap change in 7.1.0, more changes can be done later if necessary as breaking change

dakom:
Right now my setup is something like this:

```
lazy_static! {
    pub static ref JWT_ENCODING_KEY:EncodingKey = {
        let secret = std::env::var("JWT_SECRET").expect("must have JWT_SECRET set");
        EncodingKey::from_secret(secret.as_ref())
    };

    pub static ref JWT_DECODING_KEY:DecodingKey<'static> = {
        let secret = std::env::var("JWT_SECRET").expect("must have JWT_SECRET set");
        DecodingKey::from_secret(secret.as_ref())
    };
```

[...]

PR #123, opened 2020-02-27
fix issue 120: DecodingKey can be converted to static

Hexilee:
I ran into #120 as well, so did a little changes.

### DIFF
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -54,6 +54,25 @@ impl<'a> DecodingKey<'a> {
         }
     }
 
+    /// If you're using HMAC with a base64 encoded, use this.
+    pub fn from_base64_secret(secret: &str) -> Result<Self> {
+        let out = base64::decode(&secret)?;
+        Ok(DecodingKey {
+            family: AlgorithmFamily::Hmac,
+            kind: DecodingKeyKind::SecretOrDer(Cow::Owned(out)),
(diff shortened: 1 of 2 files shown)


Keats:
Thanks for that!
Can we remove the changes to 'static for now as it is a breaking change and only have the CoW + into_static? That would make it easier to use without a breaking change I think?

Hexilee:
Of course, I reverted it now.

Hexilee:
Excuse me, but, can we merge this PR and publish a new version as soon as possible? I really need it.

Keats:
@MayaWolf and @tjwilson90 would this PR work for you?

tjwilson90:
Yes, this looks good.

MayaWolf:
I went for a different approach in [my fork](https://github.com/MayaWolf/jsonwebtoken/commit/025ddc2f6cfc60dfa4df4ff15669fe7c4e0c8f39), where the DecodingKey and EncodingKey construction methods consume/move their parameters, which I think is the most versatile option - but that would be a breaking change, so if you're interested it's probably better to hold off until a major version bump.

Keats:
I think I prefer the approach of @Hexilee on that, making things Cow instead of owned so we don't **need** to clone when we can avoid it. That will be a change for the next version whenever it happens.

PR #126, opened 2020-04-21
Fix Broken chrono Example

vbrandl:
Hi,

the provided example for custom `chrono` headers (https://github.com/Keats/jsonwebtoken/blob/master/examples/custom_chrono.rs) does not work, if the `chrono` timestamp contains milliseconds. I discovered this when implementing a test property for myself that ensured, encoding and decoding results in the identity.

One if my failures looked like this:

```
[src/auth.rs:131] cookie = DeviceCookie { 
    sub: "\u{0}",                                                                    
    nonce: "\u{0}",                       
    exp: 2020-04-22T13:42:23.284112768Z,                                             
}                                    
[src/auth.rs:131] parsed = DeviceCookie {                                            
    sub: "\u{0}",                                                                    
    nonce: "\u{0}",                                                                  
    exp: 2020-04-22T13:42:23Z,     
}
```

As you can see, `cookie` (the identity) has milliseconds set but since the serializer in the example serializes only the timestamp to the second, the milliseconds get lost and the tokens are no longer the same.

I updated the example to use `Utc::timestamp_nanos` for serializing and deserializing. This solves the issue. I also added a slightly modified version of my test property to the example to ensure, that everythink works as intended.

I also fixed the existing tests since the example didn't compile and after fixing it and running `cargo test --example=custom_chrono`, the `round_trip` test fails:

```
test jwt_numeric_date::tests::round_trip ... FAILED
test jwt_numeric_date::tests::should_fail_on_invalid_timestamp ... ok
test jwt_numeric_date::tests::to_token_and_parse_equals_identity ... ok

failures:

---- jwt_numeric_date::tests::round_trip stdout ----
thread 'jwt_numeric_date::tests::round_trip' panicked at 'attempt to multiply with overflow', /home/me/.cargo/registry/src/github.com-1ecc6299db9ec823/chrono-0.4.11/src/naive/datetime.rs:342:21
```

Running the tests in release mode (`cargo test --release --example=custom_chrono`) works fine since the overflow checks are disabled for release builds.

Adding the following to `Cargo.toml` would solve the problem, too, but I don't know if you want this.

```toml
[profile.test]
overflow-checks = false
```

After disabling overflow checks for tests and running the test again, it fails with an assertion error:

```
thread 'jwt_numeric_date::tests::round_trip' panicked at 'assertion failed: `(left == right)`
  left: `"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJDdXN0b20gRGF0ZVRpbWUgc2VyL2RlIiwiaWF0IjowLCJleHAiOi00Mzg5ODA4MTQ3NDE5MTAzMjMyfQ.P2cnrIjFilrosSMBZeI5KaNDZxkxVir8l_cpvZbAP30"`,
 right: `"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJDdXN0b20gRGF0ZVRpbWUgc2VyL2RlIiwiaWF0IjowLCJleHAiOjMyNTAzNjgwMDAwfQ.RTgha0S53MjPC2pMA4e2oMzaBxSY3DMjiYR2qFfV55A"`', examples/custom_chrono.rs:63:13
```

I'm not quite sure, what to to with this test...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,8 @@ simple_asn1 = "0.4"
 [dev-dependencies]
 # For the custom chrono example
 chrono = "0.4"
+quickcheck = "0.9.2"
+quickcheck_macros = "0.9.1"
 
 [badges]
 maintenance = { status = "passively-maintained" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,6 @@ simple_asn1 = "0.4"
 [dev-dependencies]
 # For the custom chrono example
 chrono = "0.4"
-quickcheck = "0.9.2"
-quickcheck_macros = "0.9.1"
 
 [badges]
 maintenance = { status = "passively-maintained" }
--- a/examples/custom_chrono.rs
+++ b/examples/custom_chrono.rs
@@ -23,7 +23,7 @@ mod jwt_numeric_date {
     where
         S: Serializer,
     {
-        let timestamp = date.timestamp();
+        let timestamp = date.timestamp_nanos();
         serializer.serialize_i64(timestamp)
     }
 
@@ -32,9 +32,7 @@ mod jwt_numeric_date {
     where
         D: Deserializer<'de>,
     {
-        Utc.timestamp_opt(i64::deserialize(deserializer)?, 0)
-            .single() // If there are multiple or no valid DateTimes from timestamp, return None
-            .ok_or_else(|| serde::de::Error::custom("invalid Unix timestamp value"))
+        Ok(Utc.timestamp_nanos(i64::deserialize(deserializer)?))
     }
 
     #[cfg(test)]
--- a/examples/custom_chrono.rs
+++ b/examples/custom_chrono.rs
@@ -1,7 +1,3 @@
-#[cfg(test)]
-#[macro_use]
-extern crate quickcheck_macros;
-
 use chrono::prelude::*;
 use jsonwebtoken::{DecodingKey, EncodingKey, Header, Validation};
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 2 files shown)


Keats:
Hmm the example does work as intended afaik. The spec says that the times in JWTs should be UNIX timestamps (just above https://tools.ietf.org/html/rfc7519#section-30 , NumericDate). Using timestamp with milliseconds would not work in most JWTs implementation afaik as it's not part of the standard.

vbrandl:
But then using `DateTime` would be a mistake since it resolves to milliseconds.

When creating the timestamp using `Utc.now()`, it almost always includes the milliseconds. I would argue that `decode(encode(x)) == x` should always hold, which it does not in this case...

Keats:
> I would argue that decode(encode(x)) == x should always hold, which it does not in this case...

That's not possible unless you serialize in milliseconds, but then it's not a JWT anymore :/
The alternative is to ensure you don't set milliseconds in your initial dates:

Changing
```
    let mut iat = Utc::now();
    iat = iat.date().and_hms_milli(iat.hour(), iat.minute(), iat.second(), 0);
```
will make it work in round trip as you want.

vbrandl:
Sure it is possible to normalize the `DateTime` to a full second and I'm with you, that my fix should not be applied since it is against the spec, but then is it a good idea to advise users to use this kind of conversion, which might lead to bugs when the tokens are not normalized when they are created?

Keats:
We could update teh code to remove the milliseconds and add a comment explaining why this is being and whether a user *needs* to do it.

vbrandl:
I changed the example to create `Claims` with trimmed `DateTime` fields. Now `decode(encode(x)) == x` holds. I also removed the `quickcheck` dependency again and only wrote a single test.

I noticed that the `decode` function uses the current time to validate a token. Would it be possible to expose a decode function that takes `now` as a parameter? That would make testing easier since the function would be pure.

Keats:
> Would it be possible to expose a decode function that takes now as a parameter?

That would be a breaking change and make the api more verbose so no. If Rust had kwargs and default arguments on the other hand...

vbrandl:
It could be a new function `decode_at` or something like that. Or do you think that's a bad idea in general?

Keats:
I think it's a bad idea, ideally the time could be mocked instead but I don't know how mocking works in Rust :/

vbrandl:
I think in general, the rust ecosystem prefers pure functions where mocking is not needed.
The current API doesn't really allow to check if a token was valid in the past, does it?

Keats:
> The current API doesn't really allow to check if a token was valid in the past, does it?

No. You can disable the exp check and validate it manually afterwards if you want

vbrandl:
So you don't see any way to add a function to the API that allows validating at a specific point in time? Could you elaborate why you think `decode_at` would be a bad idea?

Keats:
>  Could you elaborate why you think decode_at would be a bad idea?

That offers a footgun in the API that people can easily run into and is very rarely useful. The only way I can see it making its way in this library is behind a `testing` feature disabled by default.

Veetaha:
Hey guys @Keats @vbrandl, I am wondering doesn't it make sense to add `jwt_numeric_date` module 
(with milliseconds truncation logic inbuilt) to the library itself under a feature flag so that regular users (like me :D) don't need to bother studying JWT spec on dates representation and just use what is already provided and tested.

vbrandl:
I think it would be good to hide details like this from the user, but I don't know how a good API might look. Maybe a trait `ToJwtDate` that is implemented for common date types?

PR #128, opened 2020-05-03
Decode RSA components only once

neoeinstein:
Instead of decoding the RSA modulus and exponent _every_ time that the
decoding key is required, do that once. This also ensures that any
errors with the exponent or modulus are identified early, rather than
on the first attempt to decode a token.

Note, this is a breaking change, as `from_rsa_components` can now fail
due to early validation of the modulus and exponent. Also, I've removed
the conversion through BigUInt, as performed significant work, and then
immediately undoes it. Leaving that out functions just as well.

### DIFF
--- a/src/crypto/rsa.rs
+++ b/src/crypto/rsa.rs
@@ -1,5 +1,4 @@
 use ring::{rand, signature};
-use simple_asn1::BigUint;
 
 use crate::algorithms::Algorithm;
 use crate::errors::{ErrorKind, Result};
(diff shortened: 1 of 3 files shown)


neoeinstein:
I've also moved a few of the function implementations that were producing `DecodingKey<'a>` to produce `DecodingKey<'static>` as they were producing owned `Cow`s that aren't tied to the lifetime of the arguments. This should make those functions a bit easier to use as well.

Keats:
I missed that sorry.
That's not a breaking change right?

neoeinstein:
This changeset would be breaking, as the return type of `from_rsa_components` would change from `Self` to `Result<Self>`

bdbmammoth:
Ideally, if you are gonna make this breaking change it should happen for all of the kinds of keys. As mentioned in #120. For example...

```
impl DecodingKey<'static> {
    ...
    pub fn from_ec_der(der: &[u8]) -> Result<Self> {
        Ok(DecodingKey {
            family: AlgorithmFamily::Ec,
            kind: DecodingKeyKind::SecretOrDer(Cow::Owned(der.to_vec())),
        })
    }
    etc...
}
```

Keats:
I've moved part of those changes in #160 directly.
Do we actually need the `from_rsa_raw_components`? How often do you have them in bytes rather than `&str`?

Thanks a lot @neoeinstein

PR #133, opened 2020-06-07
Updating base64 to 0.12

mettke:
Base64:
https://github.com/marshallpierce/rust-base64/compare/v0.11.0...v0.12.1
The commits show that its mostly new stuff, modularization and the switch to no_std


mettke:
I'm not completely sure why it fails but it seems that `ring v0.16.14` or `ring v0.16.13` broke compatibility with rust `1.36.0`

Keats:
Thanks!

Issue #130, opened 2020-05-17
Consider renaming dangerous_unsafe_decode


LegNeato:
Thank you for this library! Would you consider renaming `dangerous_unsafe_decode` to `dangerous_insecure_decode`? "unsafe" has a specific meaning in Rust and it might be best not to overload it (it doesn't look like the implementation uses `unsafe` at a quick glance).

If you are interested, happy to put up a patch. If not, feel free to close!

Keats:
It's unsafe because it doesn't verify the signature. Renaming it would be a breaking change though and unlikely to happen anytime soon. If you have a better name for it though, I can change it when the next major version happens.

LegNeato:
Totally get it, but it isn't unsafe in the traditional Rust memory-safety sense. My suggestion as stated above to to replace "unsafe" with "insecure". If you agree it is a change you want, I can add a function named  `dangerous_insecure_decode` and make the existing `dangerous_unsafe_decode` function `#[deprecated(note = "Please use the dangerous_insecure_decode function instead"]`.

Then you can remove at the next major rev / at your leisure.

Keats:
I like insecure!

Keats:
If anyone is up for a PR, I'll merge it otherwise that will be in a few days.

PR #139, opened 2020-06-16
Rename dangerous_unsafe_decode to dangerous_insecure_decode

giarc3:
As suggested in #130, renames `dangerous_unsafe_decode` to `dangerous_insecure_decode`. This is because "unsafe" has a very specific meaning in Rust, and this function is _not_ and `unsafe` function.

Note that this would be a breaking change and therefore require a major version bump.

### DIFF
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -184,7 +184,7 @@ pub fn decode<T: DeserializeOwned>(
 ///
 /// ```rust
 /// use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


Keats:
Can you keep the unsafe one, add a deprecated attribute and just add a insecure? version?

giarc3:
> Can you keep the unsafe one, add a deprecated attribute and just add a insecure? version?

Definitely, sorry I forgot about that suggestion. Now it won't be a breaking change.

Keats:
Thanks!

PR #141, opened 2020-06-28
Fix documentation

isolovev:


isolovev:
What is `DateTime` type? Should I indicate the package from which this type is taken?

Keats:
It should probably be a usize, it's meant to be a UTC timestamp

PR #145, opened 2020-08-06
Format code and check style in CI

piotrmaks:


Keats:
Travis is not ran anymore, I need to remove that file. It should be added to the GH actions instead.

piotrmaks:
I have reverted changes in `.travis.yml` file and enabled format check in GH actions.

Keats:
Thanks!

PR #146, opened 2020-08-17
Clippy

piotrmaks:
Enable project-wide cargo clippy and fix remaining issues

- Enable clippy in Github Actions
- Bump minimum required Rust version to 1.40.0. Motivation for this change is use of `#[non_exhaustive]` attribute that was stabilized in Rust 1.40.0
- Migrate benchmarks to criterion. Remove use of unstable features in favour of criterion benchmarks

PR #147, opened 2020-08-18
Bump minimum required Rust version to 1.40.0

piotrmaks:
Motivation for this change is use of `#[non_exhaustive]` attribute that was stabilized in Rust 1.40.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
         include:
         - build: pinned
           os: ubuntu-18.04
-          rust: 1.39.0
+          rust: 1.40.0
         - build: stable
           os: ubuntu-18.04
           rust: stable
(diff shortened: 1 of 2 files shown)


piotrmaks:
The goal is to use `#[non_exhaustive]` in #146 to fix Clippy warning that complains about the manual implementation of the non-exhaustive pattern.

Keats:
Can you group the changes in one PR? They are both fine.

piotrmaks:
Sure, I prefer small PRs but if it's fine for you I can group it. 

There is one more issue - unstable features in benchmarks cause clippy check fail on rust stable. I started working on a PR to migrate benchmarks to [criterion](https://crates.io/crates/criterion) so benchmarks can be build on stable.

 If it's ok for you I will submit all the changes in #146

piotrmaks:
I have updated #146

PR #148, opened 2020-08-31
Support more header parameters

clehner:
This adds support for the `crit` header parameter (RFC 7515), and additional header parameters via a `params` map.

I added logic for the `crit` header to check for invalid uses. Applications must do additional checking of `crit` to ensure they can understand any listed parameters.

To make the `params` map work, I removed `Hash` from the `derive` attribute of the `Header` struct. If this is a problem, I could try to find an alternative solution.

I added tests for the `crit` parameter checking and to show use of the `params` map flattened into the header struct.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -6,7 +6,7 @@ use crate::serialization::b64_decode;
 
 /// A basic JWT header, the alg defaults to HS256 and typ is automatically
 /// set to `JWT`. All the other fields are optional.
-#[derive(Debug, Clone, PartialEq, Hash, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct Header {
     /// The type of JWS: it can only be "JWT" here
     ///
@@ -42,6 +42,10 @@ pub struct Header {
     /// Defined in [RFC7515#4.1.7](https://tools.ietf.org/html/rfc7515#section-4.1.7).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub x5t: Option<String>,
+    /// Additional Public or Private header parameters
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(flatten)]
+    pub params: Option<std::collections::HashMap<String, serde_json::Value>>,
 }
 
 impl Header {
@@ -55,6 +59,7 @@ impl Header {
             kid: None,
             x5u: None,
             x5t: None,
+            params: None,
         }
     }
 
@@ -71,5 +76,23 @@ impl Default for Header {
     /// Returns a JWT header using the default Algorithm, HS256
     fn default() -> Self {
         Header::new(Algorithm::default())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+
+    #[test]
+    fn additional_parameters() {
+        let header_json = r###"{
+            "alg": "ES256",
+            "example": 123
+        }"###;
+        let header_b64 = base64::encode_config(header_json, base64::URL_SAFE_NO_PAD);
+        let res = Header::from_encoded(&header_b64);
+        let header = res.unwrap();
+        assert_eq!(header.params.unwrap().get("example").unwrap(), 123);
     }
 }
--- a/src/header.rs
+++ b/src/header.rs
@@ -42,6 +42,12 @@ pub struct Header {
     /// Defined in [RFC7515#4.1.7](https://tools.ietf.org/html/rfc7515#section-4.1.7).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub x5t: Option<String>,
+    /// X.509 certificate SHA-256 thumbprint
+    ///
+    /// Defined in [RFC7515#4.1.7](https://tools.ietf.org/html/rfc7515#section-4.1.8).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "x5t#S256")]
+    pub x5t_s256: Option<String>,
     /// Critical
     ///
     /// Defined in [RFC7515#4.1.11](https://tools.ietf.org/html/rfc7515#section-4.1.11).
@@ -64,6 +70,7 @@ impl Header {
             kid: None,
             x5u: None,
             x5t: None,
+            x5t_s256: None,
             crit: None,
             params: None,
         }
--- a/src/header.rs
+++ b/src/header.rs
@@ -1,7 +1,7 @@
 use serde::{Deserialize, Serialize};
 
 use crate::algorithms::Algorithm;
-use crate::errors::Result;
+use crate::errors::{new_error, ErrorKind, Result};
 use crate::serialization::b64_decode;
 
 /// A basic JWT header, the alg defaults to HS256 and typ is automatically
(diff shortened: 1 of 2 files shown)


Keats:
I don't mind adding that too much but is this header ever used in practice? I don't think I've ever seen anyone even mention it.

clehner:
@Keats `crit` is used for `b64` in RFC 7797 - JSON Web Signature (JWS) Unencoded Payload Option https://tools.ietf.org/html/rfc7797#section-6 - which is used in Verifiable Credentials
https://w3c.github.io/vc-data-model/#proofs-signatures

Keats:
Is that ever used in practice? Since this crate doesn't support JWS, what can you do with it?

clehner:
Hi @Keats,

I'm not sure what you mean. [RFC 7519 (JWT)](https://tools.ietf.org/html/rfc7519) says a JWT is either a JWS or a JWE. This crate implements JWT as JWS. RFC 7519 specifies that header parameters apply from [RFC 7515 (JWS)](https://tools.ietf.org/html/rfc7515) or [RFC 7516 (JWE)](https://tools.ietf.org/html/rfc7516), depending on whether the JWT is a JWS or JWE.

The `crit` header is also defined for JWE, with the same meaning as in JWS - similarly to the other currently implemented parameters.

I am looking to use JWT with `crit` and `b64` header params for Verifiable Credentials & [Linked Data Proofs](https://w3c-ccg.github.io/ld-proofs/).

Keats:
Sorry, I meant that decoding a JWT with `crit = [b64]` is not going to work with this library. Is the idea to not validate it?

clehner:
@Keats
The idea with `crit` here is that the application (library user) validates it - beyond the validation added in this PR. An alternative would be if the application somehow specifies to the library which params it understands and then we use that to validate `crit`.

`b64: false` I am seeing now will be harder to use with this library, as even if the JWK is manually encoded/decoded, the sign and verify functions in `crypto` deal with string slices rather than bytes. I will open another issue for that.

Keats:
Thanks for opening #149 
I don't know how to handle that decoding to be honest, it is very niche so I don't want to change the existing API for 99.9% of the users.

clehner:
@Keats would you be open to a PR adding support if it did not break the existing API? For example, adding sign and verify methods for byte slices rather than strings. Or do you mean you want no changes to the API for these cases, even non-breaking changes?

Keats:
Non-breaking changes are fine. The change I had in mind were more on how to skip the b64 step in the example of `b64: false`

clehner:
@Keats thanks for merging #150 and #154.

I've rebased this PR onto `next` ([v8](https://github.com/Keats/jsonwebtoken/pull/160)), and added support for the `x5t_s256` header parameter. I am not using the `x5t_s256` parameter; it is just for completeness. There is also the `jwk` parameter that could be added, but I did not add that since there is already a PR open for it in #119, and it is not clear to me what the type should be for it anyway.

If you prefer not to include the `x5t_s256` and `crit` headers, please consider just the first commit in this PR. That adds a `params` property flattened into the Header struct. This would be a "catch-all" for additional arbitrary properties that library users could use, which would otherwise be discarded during deserialization. This would address #119 as well. But note that once the `params` property is added, adding additional properties would be a breaking change, since they would then no longer be accessible via the `params` property.

Keats:
No problem, sorry for taking so long to get back to you.
I think this is fine, maybe change the name of `params` to something like `additional_headers` but I'll have another look tomorrow when I'm a bit more rested :)

src/header.rs
```diff
@@ -42,6 +42,21 @@ pub struct Header {
...
+    pub x5t_s256: Option<String>,
+    /// Critical
+    ///
+    /// Defined in [RFC7515#4.1.11](https://tools.ietf.org/html/rfc7515#section-4.1.11).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub crit: Option<Vec<String>>,
+    /// Additional Public or Private header parameters
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(flatten)]
+    pub params: Option<std::collections::HashMap<String, serde_json::Value>>,
```
Keats, line 59: I'm a bit torn about exposing `Value` here... Can we not assume that it's going to be 99% Strings or are there known usecases where it's not a string? Outside of some homegrown token implementation

It looks like we're missing `x5c` too? I need to go through the spec after merging and ensure we have everything

src/header.rs
```diff
@@ -55,15 +70,40 @@ impl Header {
...
+        if let Some(crit) = &header.crit {
+            if crit.is_empty() {
+                return Err(new_error(ErrorKind::InvalidCriticalHeader));
+            }
+            for name in crit {
+                match name.as_str() {
+                    "alg" | "jku" | "jwk" | "kid" | "x5u" | "x5c" | "x5t" | "x5t#S256" | "typ"
+                    | "cty" | "crit" => return Err(new_error(ErrorKind::InvalidCriticalHeader)),
+                    _ => {
+                        let has = match &header.params {
```
Keats, line 94: has -> has_param?

Keats (changes requested):

src/header.rs
clehner: Yes, I forgot `x5c`. [RFC 7515 § 4.1.6](https://tools.ietf.org/html/rfc7515#section-4.1.6) specifies `x5c` as an array of strings (in base64-encoding).

[§ 4.1.11](https://tools.ietf.org/html/rfc7515#section-4.1.11) gives a hypothetical example of a header value that is a number.

[§ 4.1.3](https://tools.ietf.org/html/rfc7515#section-4.1.3) specifies `jwk` is a JWK, which is an object, specified in [RFC 7517](https://tools.ietf.org/html/rfc7517) with its own [registry](https://www.iana.org/assignments/jose/jose.xhtml#web-key-parameters) of parameters - some of which appear to be the same as JOSE header parameters, and several various key parameters.

I am using the `b64` header option (as in #149), which is a boolean.

`crit` is an array of strings.

In the [JWS/JWE header parameters registry](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-header-parameters), the remaining parameters for JWS not already mentioned are:
- `url` and `nonce`, from [RFC 8555](https://tools.ietf.org/html/rfc8555) (ACME): String
- `ppt`: has its own [registry](https://www.iana.org/assignments/passport/passport.xhtml#passport-extensions) of possible values, which are all strings

I can understand if you don't want to expose Value for arbitrary parameters, as it looks like maybe not the fullest use of the type system. On the other hand, [RFC 7515 § 4.2](https://tools.ietf.org/html/rfc7515#section-4.2) and [§ 4.3](https://tools.ietf.org/html/rfc7515#section-4.3) say that additional header values may be used that are not in the registry (with caution / sparingly), and doesn't place any type restrictions on them. So it would be adhering to the RFC to allow additional arbitrary headers.  With `Value`, a library user could use `serde_json::from_value`/`to_value` to do additional (de)serialization.

If the additonal headers map is not included, would you be open to instead including specific support for additional standardized header? Specifically, I would add `crit` as in this PR, and `b64` (which is easier now that we can sign bytes).

Keats: I'm thinking adding additional headers should be via a method on `Header` rather than making the field pub. This way we don't need to expose Value, we can just take something that `impl dyn Serialize`. We also need to add another method that returns the HashMap, which would expose Value here, no way around it. But at least we got the serialization taken care of by the library.

clehner:
@Keats Sorry for letting this sit; I am closing this PR now as I have addressed my use case a different way.

Issue #149, opened 2020-09-02
Support Unencoded Payload Option


clehner:
[RFC 7797 - JSON Web Signature (JWS) Unencoded Payload Option](https://tools.ietf.org/html/rfc7797)

RFC 7797 adds a JWT header option `"b64": false` which specifies that the payload is not base64url-encoded but is added to the signing input as bytes:

```
   +-------+-----------------------------------------------------------+
   | "b64" | JWS Signing Input Formula                                 |
   +-------+-----------------------------------------------------------+
   | true  | ASCII(BASE64URL(UTF8(JWS Protected Header)) || '.' ||     |
   |       | BASE64URL(JWS Payload))                                   |
   |       |                                                           |
   | false | ASCII(BASE64URL(UTF8(JWS Protected Header)) || '.') ||    |
   |       | JWS Payload                                               |
   +-------+-----------------------------------------------------------+
```

This is useful with [Detached Content](https://tools.ietf.org/html/rfc7515#appendix-F) where the JWT is serialized without the payload.

PR #150, opened 2020-09-08
Add sign and verify on bytes

clehner:
Re: https://github.com/Keats/jsonwebtoken/pull/148#issuecomment-689074489

This adds `sign_bytes` and `verify_bytes` functions in `crypto`, corresponding to the existing public `sign` and `verify` functions but accepting a message as bytes instead of as a string. The existing `sign` and `verify` functions are changed to be implemented in terms of the new `sign_bytes` and `verify_bytes` functions.

I changed the internal sign/verify functions to use bytes instead of strings rather than adding new functions there too, since those functions are not exported.

Signing and verifying messages as bytes instead of (base64url-encoded) strings is needed for the Unencoded Payload Option (#149).


Keats:
I think passing bytes everywhere is more flexible anyway and there is an upcoming major version so we can have breaking changes.

PR #154, opened 2020-10-15
Add EdDSA (Ed25519)

clehner:
This adds support for Ed25519 keys, from the EdDSA family.

References:
- [RFC 8037 - CFRG ECDH and Signatures in JOSE](https://tools.ietf.org/html/rfc8037)
- [RFC 8410 - Safe Curves for X.509](https://www.rfc-editor.org/rfc/rfc8410.html)

RFC 8037 also defines Ed448, but `ring` doesn't have that (https://github.com/briansmith/ring/issues/463), so we don't use it.

I didn't add a test using jwt.io, as it doesn't look like they support EdDSA.


Keats (approved):
Looks good thanks!

Issue #156, opened 2020-11-04
Decode arbitrary payload


superstator:
I'd like to use `jsonwebtoken` to decode a JWT with an arbitrary payload, and just return that payload as a string. I still want to verify the signature, but I don't want to look for keys like `iat`, `exp`, etc. In fact I'd like to handle cases where the payload isn't JSON at all, or may be malformed JSON. Is this possible? If not, would you be open to a PR?

Keats:
In that case it's easier to skip this crate and just split on `.` and decode base64

superstator:
But I do need signature verification, and just I don't want to actually decode/deserialize anything. You have a `verify` fn in `/src/crypto/mod.rs` that does most of what I need, though it would be nice to also use your code for error handling and splitting up the JWT compact string. Basically, everything in `jsonwebtoken::decode` but the last couple lines, call it something like `jsonwebtoken::verify_sig`.

I actually have it working in a fork, happy to send a PR or a link, if that would make it clearer what I'm trying to do.

Keats:
Ok, I'll take a PR

PR #161, opened 2020-11-18
add verify_sig fn

superstator:
For #156. I'm essentially just pulling the signature verification steps into their own function that can be called independently, and then having `decode` call that new function rather than duplicating everything. I'm certainly not set on `verify_sig` as a function name, if you have suggestions.

### DIFF
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -131,6 +131,43 @@ impl<'a> DecodingKey<'a> {
     }
 }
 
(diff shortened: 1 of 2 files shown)


Keats:
Can you base your PR on https://github.com/Keats/jsonwebtoken/pull/160 ?
There are some changes around those functions in it.

superstator:
Like that? Or would you rather I actually merge `next` on my end?

PR #163, opened 2020-12-06
Add note about encoding keys

arniu:


src/encoding.rs
```diff
@@ -29,6 +29,16 @@ impl EncodingKey {
...
     /// This errors if the key is not a valid RSA key.
+    ///
+    /// # NOTE
+    ///
+    /// According to the [ring doc](https://briansmith.org/rustdoc/ring/signature/struct.RsaKeyPair.html#method.from_pkcs8),
+    /// the key should be at least 2047 bits.
+    ///
+    /// You can generate a key with the following:
+    ///
+    ///     openssl genrsa -out rsa-private.pem 2048
```
Keats, line 40: I wouldn't put the openssl commands in the docs, it's already mentionned in the ring docs

Keats (changes requested):

Keats:
Thanks!

PR #167, opened 2021-01-09
Update ring to 0.16.9

andrekorol:
Allow crates that depend on jsonwebtoken to compile successfully on Apple Silicon (aarch64-apple-darwin).


Keats:
That doesn't change anything due to how cargo works, 0.16.9 will get picked up automatically if you `cargo update` in your project.

PR #168, opened 2021-01-20
Makes Algorithm traits available for library clients

rlcintra:
The FromStr implementation for Algorithm is useful when clients of the
library are dealing with keys not provided at compile time (e.g. keys
retrieved from OIDC). Making the algorithm module public exposes the
FromStr and the Display traits for clients. This change is backward
compatible.


rlcintra:
Apologies, the traits are already available without needing to make the `algorithms` module public.

PR #171, opened 2021-02-19
Bump dependencies (base64, simple_asn1), clippy lints

adamchalmers:
My project uses your library, thank you for writing it, it's very useful. I was going through my project, cleaning up old dependencies and looking for any projects that pulled in different versions, and upgrading them to the newer version where possible. Seems like these version bumps don't involve any breaking changes, as your library still compiles and passes its tests after bumping the version. So I went ahead and opened this PR. 

I initially just wanted to upgrade the dependencies. But then I noticed CI was failing due to the new Clippy lints in new Rust releases, so I fixed those too. 

It'd be great if you could cut a new v7.2.1 release after merging this! That way I wouldn't need to wait for the v8 you're working on to be finished. Thank you for your time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,10 +15,10 @@ include = ["src/**/*", "LICENSE", "README.md", "CHANGELOG.md"]
 serde_json = "1.0"
 serde = {version = "1.0", features = ["derive"] }
 ring = { version = "0.16.5", features = ["std"] }
-base64 = "0.12"
+base64 = "0.13"
 # For PEM decoding
 pem = "0.8"
-simple_asn1 = "0.4"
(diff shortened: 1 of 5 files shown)


Keats:
> That way I wouldn't need to wait for the v8 you're working on to be finished. Thank you for your time.

I've actually just released v8 beta and don't expect any big changes API wise. I would recommend trying that instead for the time being.

adamchalmers:
OK, if you're already on v8 beta then I won't worry about cutting 7.2.1. I'd be happy if these changes were included in the v8 release!

Keats:
It's already included ;)

PR #172, opened 2021-02-21
Multiple iss field validation check

sabify:
Hello,

Some providers like [Google](https://developers.google.com/identity/sign-in/web/backend-auth) need to validate `iss` field against multiple values. It would be better to have `HashSet` type for `iss` field to validate multiple `iss` values instead of using multiple validation objects and multiple validations.


Keats:
Can you do a PR with a hashset on https://github.com/Keats/jsonwebtoken/pull/160 ?

sabify:
Done. #173

PR #174, opened 2021-02-25
Updated base64 to v0.13 and simple_asn1 to v0.5

lpil:
Hello! Thank you for your library, we have found it tremendously useful in our Rust video game projects over at Embark Studios.

We use [cargo-deny](https://github.com/EmbarkStudios/cargo-deny) to ensure we don't have multiple versions of the same deps, and it notified us that there was a duplicate resulting from this library using an older version.
This PR updates those deps to the latest version.

Thank you,
Louis

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+- Updated `base64` to v0.13
+- Updated `simple_asn1` to v0.5
+
 ## 7.2.0 (2020-06-30)
 
 - Add `dangerous_insecure_decode` to replace `dangerous_unsafe_decode`, which is now deprecated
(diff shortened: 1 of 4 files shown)


lpil:
I think CI has failed due to a network problem:

```
error: could not download file from 'https://static.rust-lang.org/rustup/release-stable.toml' to '/tmp/rustup-updated9FDFl/release-stable.toml'
error: caused by: failed to make network request
error: caused by: error sending request for url (https://static.rust-lang.org/rustup/release-stable.toml): connection error: Connection reset by peer (os error 104)
error: caused by: connection error: Connection reset by peer (os error 104)
error: caused by: Connection reset by peer (os error 104)
```

Keats:
Check the v8 beta on crates.io it already includes all those changes + some UX improvements
It's very close to release, I was just hoping to get some more feedback.

lpil:
Oh wonderful, thank you.

PR #175, opened 2021-03-02
Validation of algorithms must be skipped if algorithms are empty

olitha:
Documentation of validation algorithm member states:
https://github.com/Keats/jsonwebtoken/blob/2f25cbed0a906e091a278c10eeb6cc1cf30dc24a/src/validation.rs#L62-L63

*Expected*
no validation if empty Vec set.

*Observed*
Empty Vec always returns error, because contains can never succeed.

### DIFF
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -290,3 +290,40 @@ pub fn decode_header(token: &str) -> Result<Header> {
     let (_, header) = expect_two!(message.rsplitn(2, '.'));
     Header::from_encoded(header)
 }
+
+#[cfg(test)]
+mod test {
+    use super::{verify_signature, DecodingKey, Validation};
+    use crate::Algorithm;
+
+    #[test]
+    pub fn validate_alg_too_strict_fails_with_correct_token() {
+        let validation = Validation {
+            algorithms: vec![Algorithm::ES256, Algorithm::RS256],
+            ..Validation::default()
+        };
+        let key_pem = "-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE1n8lxw87OQh39IJYRNRKl3eoxhjf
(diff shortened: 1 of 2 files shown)


Keats:
Sorry I moved the PR to the `next` branch and it messed up things a bit.
I believe we should move that check before the validation even, it should be the first thing we do when decoding so we don't do any work if it's empty

olitha:
No Problem. i didn't know from which branch to start :) I merged it with next and resolved the conflict
About the validation, I think only the validation of algorithms should be skipped, not the whole, right?

Keats:
So my thinking is: should we even allow Validation to not have an algorithm set? It sounds like a recipe for disasters to not check an `alg` so we should enforce that there is one when decoding. It doesn't make sense *not* validating which alg was used, if you don't know it there is https://docs.rs/jsonwebtoken/7.2.0/jsonwebtoken/fn.decode_header.html to get it.

olitha:
I totally messed up with rebasing... Is it possible to reopen it?

Keats:
No commits in it so I can't re-open. Just submit a new one

olitha:
To me it looks like one could just use the verify_signature fn with any key type in principle and validation of alg is too strict.
Does it make sense to validate the key family here?
https://github.com/Keats/jsonwebtoken/blob/2f25cbed0a906e091a278c10eeb6cc1cf30dc24a/src/decoding.rs#L157-L159
I created a failing and succeeding test for it to show.

Additionally as far as I understand, even if the key would be of the same family, say Ec, if I set the decoding key to a P-384 and alg says ES256 (P-256) this would pass validation, but fail verification.

So I would suggest to skip the validation of family there and just let the ciphers pass, that are set in Validation. This validation could be moved to the validate fn.

Does this make sense?

Keats:
I think this check is there just to fail early if someone passes the wrong thing - no need to bother doing any work if that's the case. Ideally this would be a compile-time check but it's not: we are just validating the user provided variables, not the JWT itself.
I think it's different than actually validating the alg from the header

Keats:
Ok I've added that check in the next release branch.

PR #176, opened 2021-03-03
increasing performance over validation of audiences

sabify:
Hello,

We don't need to iterate over all possible hits in two sets by counting them, we just need the first hit to validate.

### DIFF
--- a/src/validation.rs
+++ b/src/validation.rs
@@ -159,7 +159,7 @@ pub fn validate(claims: &Map<String, Value>, options: &Validation) -> Result<()>
                 }
                 Value::Array(_) => {
                     let provided_aud: HashSet<String> = from_value(aud.clone())?;
-                    if provided_aud.intersection(correct_aud).count() == 0 {
+                    if provided_aud.intersection(correct_aud).next().is_none() {
                         return Err(new_error(ErrorKind::InvalidAudience));
                     }
                 }


src/validation.rs
```diff
@@ -159,7 +159,7 @@ pub fn validate(claims: &Map<String, Value>, options: &Validation) -> Result<()>
                 }
                 Value::Array(_) => {
                     let provided_aud: HashSet<String> = from_value(aud.clone())?;
-                    if provided_aud.intersection(correct_aud).count() == 0 {
+                    if provided_aud.intersection(correct_aud).next().is_none() {
```
Keats, line 162: Does that make any difference in practice when the size of each set will be <2-3? I would go for the clearer approach with count if that's not the case

sabify:
Here is the difference:

It's about 2.5x performance gain.

```sh
bench_hashmap_count     time:   [43.185 ns 43.226 ns 43.271 ns]
Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) high mild
  3 (3.00%) high severe

bench_hashmap_next      time:   [16.125 ns 16.128 ns 16.133 ns]
Found 13 outliers among 100 measurements (13.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild
  8 (8.00%) high severe
```

Code:
```rust
fn bench_hashmap(c: &mut Criterion) {
    let auds1: std::collections::HashSet<String> =
        ["1".to_string(), "2".to_string(), "3".to_string()].iter().cloned().collect();
    let auds2: std::collections::HashSet<String> =
        ["1".to_string(), "2".to_string(), "3".to_string()].iter().cloned().collect();
    c.bench_function("bench_hashmap_count", |b| b.iter(|| auds1.intersection(&auds2).count()));
    c.bench_function("bench_hashmap_next", |b| {
        b.iter(|| auds1.intersection(&auds2).next().is_none())
    });
}
```

Keats:
I meant across the whole decoding process, I'm guessing it's negligeable

sabify:
But affect the performance a lot in its place! specially on validating multiple validation audiences over multiple claim audiences.
I'm still don't get it you've said `I would go for the clearer approach with count...`
Why `.next().is_none()` is not clear but `.count() == 0` is?

Keats:
> Why .next().is_none() is not clear approach but .count() == 0 is?

Probably just a personal thing. I'm happy to merge it if it's not a common viewpoint.

> But affect the performance a lot in its place! specially on validating multiple validation audiences over multiple claim audiences.

Over the whole decoding, it's just a rounding error. 20ns out of 2800ns for a whole decoding on my machine

PR #177, opened 2021-03-04
FEATURE: A Validator Struct Bridge

sabify:
Would it be more convenient to have a bridge structure to store `DecodingKey` and `Validation` to validate tokens/claims?


Keats:
I don't think this is bringing any value, you can just create a Validation object next to the DecodingKey and use it on a `decode` without adding any layer. The `src/validation.rs` should be its own PR though.

Issue #181, opened 2021-03-05
Add x5c Header


olitha:
The x5c Header contains a certificate chain as base64 encoded DER certificates and can be used outside the jsonwebtoken lib to get the public key for signature verification.
Additionally the certificate chain can be verified by the application.

Can we add the x5c header? Only to make it available to applications?

jnicholls:
Wow, I literally implemented this last night with the intention to open an issue and submit a PR this morning. However, I did it differently and I think more correctly. I will submit my PR shortly after I rebase with your update.

jnicholls:
@Keats Do you like to close out these issues once a release is done, or after code is in `next`? I'll quickly take the opportunity to say thanks for your contributions to the community as well!

Keats:
Forgot to reply! Thanks for the kinds words, I'll close all relevant issues on release - probably this weekend if nothing else pops up.

PR #184, opened 2021-03-05
Add an access method to decode the Header x5c field into DER PKIX format.

jnicholls:
Another take on #181 that properly base64-decodes the X.509 certificate chain.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -1,3 +1,5 @@
+use std::result;
+
 use serde::{Deserialize, Serialize};
 
 use crate::algorithms::Algorithm;
@@ -39,7 +41,7 @@ pub struct Header {
     pub x5u: Option<String>,
     /// X.509 certificate chain. A Vec of base64 encoded ASN.1 DER certificates.
     ///
-    /// Defined in [RFC7515#](https://tools.ietf.org/html/rfc7515#section-4.1.6).
+    /// Defined in [RFC7515#4.1.6](https://tools.ietf.org/html/rfc7515#section-4.1.6).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub x5c: Option<Vec<String>>,
     /// X.509 certificate thumbprint
@@ -59,8 +61,8 @@ impl Header {
             jku: None,
             kid: None,
             x5u: None,
-            x5t: None,
             x5c: None,
+            x5t: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


Keats:
> Ensures the new x5c field, as well as all the other optional RFC7515 fields, are able to be missing when deserializing a Header.

That's already the case...? See all the tests and pretty much every user of the crate

jnicholls:
Yes I was being overly explicit, sorry. The MissingFieldDeserializer handles the Option case already. The reason I thought to add it is when you provide your own `deserialize_with` implementation, the MissingFieldDeserializer is no longer used in the `serde_derive` generated code, so I had to add `#[serde(default)]` explicitly.

I had already worked to remove that stuff as you pointed this out, it's updated now.

src/header.rs
```diff
@@ -1,9 +1,13 @@
-use serde::{Deserialize, Serialize};
+use std::result;
+
+use serde::{de, Deserialize, Serialize};
 
 use crate::algorithms::Algorithm;
 use crate::errors::Result;
 use crate::serialization::b64_decode;
 
+type X509Chain = Vec<Vec<u8>>;
```
Keats, line 9: Doesn't this type need to be public? I think I would rather skip it tbh as it's only used in this file and the actual type is clearer

Keats (changes requested):

src/header.rs
jnicholls: The public-with-private type checker automatically replaces the type with private type aliases in these types of cases when the aliased type is public. I removed it.

olitha:
I wonder if it Is necessary to decode base64 when deserializing. There are other fields already like x5t which is base64 (this time base64url) and is not decoded by serde. On the one hand it is good to hide these details (the exact coding) away from the user, on the other hand there is processing overhead even for fields that are not used by application code.

As an idea. What about classical `impl Header` with getters that decode? This would remove processing *and* hide the encoding details. But would also leave the user being able to get the original values.

jnicholls:
@olitha I tend to agree, albeit there is still deserialization overhead w/ the allocation of the intermediate Strings, just not nearly as much as what base64 decoding introduces, particularly if the application wants to make use of the field as it will always want the DER representation for any practical X.509 use case. The base64 decoder looks to be quite efficient (several GB/s on commodity hardware). Nevertheless, the best argument is to pay the cost only when it's needed, so I will revise.

Keats:
I like that!

jnicholls:
Yes I will add a test. Is in header.rs okay?

Keats:
yep

jnicholls:
On second thought, I might put them into a `tests/header.rs` module. I want to have a legit certificate `.der` file to include into one of the tests, rather than put it into the code directly. I'll move in that direction, but let me know if you have any objections to that.

PR #185, opened 2021-03-06
Add tests for Header::x5c_der(), and modify its return type to be fallible.

jnicholls:
Add some tests to cover the new `Header::x5c_der()` method added in #184.


tests/header/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+
+#[test]
+fn x5c_der_invalid_chain() {
+    let mut x5c: Vec<_> = CERT_CHAIN.iter().map(ToString::to_string).collect();
+    x5c.push("invalid base64 data".to_string());
+
+    let x5c = Some(x5c);
+    let header = Header { x5c, ..Default::default() };
+
+    assert!(header.x5c_der().is_none());
```
Keats, line 33: should it return an error instead of an option?

jnicholls: Well here are some of our options (no pun intended):

1. Convert `Option<Vec<String>>` to `Option<Vec<Vec<u8>>>` and return `None` if any decoding error occurs (current implementation).
2. Convert `Option<Vec<String>>` to `Option<Vec<Vec<u8>>>` and drop any certs do not decode successfully, e.g. using `filter_map`.
3. Convert `Option<Vec<String>>` to `Result<Vec<Vec<u8>>, _>` and return an Error if the field is `None` or any decoding errors occur. This is semantically equivalent to #1 and thus not a novel option.
4. Convert `Option<Vec<String>>` to `Result<Option<Vec<Vec<u8>>>, _>` and return an Error if any decoding errors occur. This distinguishes between a decoding failure and the original field being `None`.
5. Convert `Option<Vec<String>>` to `Option<(Vec<Vec<u8>>, Vec<String>)>` and return a tuple of certs in-order that successfully decoded, and a list of certs in-order that failed to decode. This does not block the caller from getting access to some of the decoded certs, and it collects the corrupt certs as an aside for inspection.

#4 is semantically better than the current implementation (#1/#3), but it's still an all-or-nothing result. #2 is infallible but lossy. #5 is infallible and lossless.

Thoughts?

Keats: I would go for the option 4. If someone really wants to find out the valid certs in the list they can do it themselves via the field of the struct.

jnicholls: Ok wilco. I'll update this PR to incorporate the API change and the tests to go along with it.

Keats (approved):

Keats:
Thanks!

PR #195, opened 2021-07-14
Add types for JWK/JWKS based on biscuit

Keats:


Cargo.toml
```diff
@@ -7,7 +7,7 @@ readme = "README.md"
 description = "Create and decode JWTs in a strongly typed way."
 homepage = "https://github.com/Keats/jsonwebtoken"
 repository = "https://github.com/Keats/jsonwebtoken"
-keywords = ["jwt", "web", "api", "token", "jwk"]
+keywords = ["jwt", "web", "api", "token", "jwk", "jwks"]
```
a1ien, line 10: https://doc.rust-lang.org/cargo/reference/manifest.html#the-keywords-field

> Note: crates.io has a maximum of 5 keywords. Each keyword must be ASCII text, start with a letter, and only contain letters, numbers, _ or -, and have at most 20 characters.

Keats: I keep forgetting about that :(

Cargo.toml
```diff
@@ -24,6 +24,9 @@ simple_asn1 = "0.5"
 # For the custom chrono example
 chrono = "0.4"
 criterion = "0.3"
+# for the auth0 JWK example
+reqwest = { version = "0.11", features = ["json"] }
```
btielen, line 28: Not being used in the example

examples/auth0.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    let header = decode_header(TOKEN)?;
+    let kid = match header.kid {
+        Some(k) => k,
+        None => return Err("Token doesn't have a `kid` header field".into()),
+    };
+    match jwks.find(&kid) {
+        Some(j) => {
+            let alg = j.common.algorithm.unwrap();
+            match j.algorithm {
```
btielen, line 24: A match inside a match makes it difficult for me

btielen (commented):
Just two things I noticed. I could take a deeper look in the example, because I have experience with Auth0

PR #198, opened 2021-08-04
Fixing typo in doc comment

Luca-spopo:
The doc comment for `InvalidSubject` is accidentally the same as the doc comment for `InvalidAudience`


Keats:
Thanks, this is already fixed in the upcoming version so closing it.

PR #202, opened 2021-08-25
Improve deserialization performance around validation and its tests

Ten0:
The claims validation was done via deserializing into a Map, which
implies allocations/deallocations. This was done even if the map was not
used afterwards.

This commit improves performance of the validation by never
deserializing in a `Map`, and deserializing only when necessary, to
a struct that typically only borrows from the original b64-decoded
json string.

The validation function interface change required update to the tests,
which are also made easier to read by using the `serde_json::json!`
macro.


Keats:
@Ten0 sorry for the delay, can you rebase?

Ten0:
> @Ten0 sorry for the delay, can you rebase?

Done.
Considering that you're squash-and-merging anyway I favored merging to keep the history in this PR.

Keats:
Perfect, thanks!

PR #207, opened 2021-09-15
Update base64 to 0.13 and simple_asn1 to 0.5

LokiVKlokeNaAndoke:
Seems like nothing got broken


Keats:
The v8 beta already has that in it.

PR #213, opened 2021-10-29
(`next`) Remove chrono

Adam-Gleave:
Removes `chrono` as a dependency of the `next` branch. See #211.


Keats:
It looks the the pinned version needs to be bumped to 1.51

PR #215, opened 2021-10-29
Update  `simple_asn1` to `0.5`

ereOn:
This PR bumps the `simple_asn1` crate to its greatest and latest version, so that people can keep using `jsonwebtoken` in their projects.


Keats:
Thanks, already covered in https://github.com/Keats/jsonwebtoken/pull/213

PR #218, opened 2021-11-16
Implement Clone, Eq and PartialEq for Error.

### LINKED ISSUE #217: Implement Clone and Eq for error type
It would be helpful if `jsonwebtoken::error::Error` implemented `Clone` and `Eq`.

qwandor:
This fixes #217.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -11,7 +11,7 @@ pub(crate) fn new_error(kind: ErrorKind) -> Error {
 pub type Result<T> = result::Result<T, Error>;
 
 /// An error that can occur when encoding/decoding JWTs
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 pub struct Error(Box<ErrorKind>);
 
 impl Error {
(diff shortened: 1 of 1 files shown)


Keats:
Can I just get more details on when you would use eq on the error instead of on the error kind? And when is PartialEq not enough?

qwandor:
I'd like to be able to `#[derive(Clone, Eq, PartialEq)]` my own error types which wrap `jsonwebtoken::Error`. In particular I have something like this:

```rust
#[error(Debug, Clone, Eq, PartialEq, thiserror::Error)]
pub enum MyError {
    #[error("JWT error: {0}")]
    Jwt(jsonwebtokenerror::Error),
    #[error("Some other error...")]
    SomeOtherError(...),
    ...
}
```

For the derive macros to work, the various wrapped error types need to implement `Clone` and `Eq` themselves. All the error types provided by other libraries I use do, it's just `jsonwebtoken` which doesn't.

As for `PartialEq` vs. `Eq`, I guess `PartialEq` is 'enough' in that it implements all the methods I need, but there's no reason not to implement `Eq` and it's correct to do so. It's just a marker trait which says that your `PartialEq` implementation is reflexive, symmetric and transitive, which it already is.

PR #221, opened 2021-11-19
Upgrade base64 to v0.13

giarc3:
Upgrades to the latest version of base64
Unfortunately, this could be considered a breaking change because `base64::DecodeError` is part of the public API as a variant of `ErrorKind`. But I will leave that decision up to the maintainers.


Keats:
This is already done on the beta version

PR #222, opened 2021-11-19
Upgrade from simple_asn1 0.4 -> 0.6 to address two RUSTSEC advisories

clintfred:
This removed a transitive dependency on `chrono` and upgrades a dependency on `time` to address the following security vulnerabilities. 

- https://rustsec.org/advisories/RUSTSEC-2020-0159.html
- https://rustsec.org/advisories/RUSTSEC-2020-0071.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ ring = { version = "0.16.5", features = ["std"] }
 base64 = "0.12"
 # For PEM decoding
 pem = "0.8"
-simple_asn1 = "0.4"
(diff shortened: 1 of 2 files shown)


clintfred:
I see that this is already on https://github.com/Keats/jsonwebtoken/tree/next, but perhaps it would be good to release this as a patch release as well. If you want to wait until 8 to address this, feel free to close this PR.

Keats:
At this point it's better to use the beta, it's 99% done and hopefully get some more feedback on it before releasing v8

PR #224, opened 2021-11-27
Update simple asn1

mathstuf:
This drops a dependency on `chrono` which, by default, depends on `time 0.1` through its `oldtime` feature. This version of `time` has a security vulnerability, so bump to avoid adding the crate to dependency trees.


Keats:
This is already done on the v8 beta.

Issue #190, opened 2021-06-19
The exp claim should be optional


cloutiertyler:
According to the RFC, the `exp` claim should be optional: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4

> 4.1.4.  "exp" (Expiration Time) Claim
> ... Use of this claim is OPTIONAL.

I saw that you suggested that the user just set an expiration very far in the future. This works for new tokens, however it doesn't work for tokens created previously (e.g. if you're migrating tokens to a new service).

Keats:
We could make it optional but is there anyone using JWTs without exp?

cloutiertyler:
We were issuing some tokens without `exp` on our old service because they were stored securely and we can also manually expire them, but regardless of that I think the library should implement the RFC to spec. In my experience, even small deviations from the RFC can have unintended consequences that aren't immediately obvious (migrating from an old service is just one example).

Fortunately in this case the RFC isn't ambiguous at all and since it's going from required to optional it's a non-breaking change for the library. I can definitely understand the desire for wanting `exp` to be mandatory, since it's good practice, but I don't think it's worth potentially preventing someone from using the library because their systems rely on it's optionality.

Maybe there's a way for the library to encourage its use, without absolutely requiring it.

Keats:
> Maybe there's a way for the library to encourage its use, without absolutely requiring it.

Keats:
Actually looking at it again, if you set `Validation.validate_exp` to `false` it should be working fine?

brocaar:
@Keats I think I'm running into the same issue as @cloutiertyler. I don't think the `validate_exp = false` is the solution. 

In my use-case, two different types of claims can be issued, claim A will have an `exp` set (and must be validated), claim B does not have an `exp` set (and there is no expiration to validate).

Currently this is not possible as `validate_exp = true` will invalidate claim B (as no `exp` is considered an expired claim if I'm not mistaken), `validate_exp = false` is not an option as claim A with an `exp` must be validated. You only know which claim was received after validating it, so you can't adapt your validation settings per claim type.

My thoughts:

Is there an use-case for the `validate_exp` option? In other words, if you have a claim with an `exp` which has already expired, is there an use-case to let it pass the validation?

By removing the `validate_exp` option and only validating the `exp` if it is set, users still have the flexibility to use it or not, based on if they include the `exp` into the generated claims or not (which makes it optional to use). I think the same applies to `validate_nbf`.

[...]

PR #225, opened 2021-12-03
Add required_spec_claims

Keats:


brocaar:
Thanks @Keats. If I read the changes correctly, It means this solves #190 by not including `exp` in `required_spec_claims` (which will validate the presence of the claims) and by setting `validate_exp` to `true` (which validates the `exp` only if it is included in the claim) :+1:

Keats:
Yep exactly

PR #230, opened 2022-01-29
Lifetime introduction for 0 copy keys proposal

BRA1L0R:
EndodingKey and DecodingKey do unecessary copies of secret keys or ders when borrowing could be an option.

This is a proposal because I can still see scenarios where a user would want a completely owned key, but if this is the case it should be documented and specified that secrets are always cloned.

### DIFF
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use serde::de::DeserializeOwned;
 
 use crate::algorithms::AlgorithmFamily;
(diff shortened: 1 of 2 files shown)


Keats:
So there were actually issues with using lifetimes with them before: https://github.com/Keats/jsonwebtoken/issues/120
I've moved to owning the key entirely since it's usually not something you call often - most of the time you instantiate your keys once.

BRA1L0R:
> So there were actually issues with using lifetimes with them before: #120 I've moved to owning the key entirely since it's usually not something you call often - most of the time you instantiate your keys once.

Ok it makes sense it would indeed incentivize reusing keys instead of creating them multiple times.

PR #232, opened 2022-01-29
Reduced default double allocation to one

BRA1L0R:
Default previously allocated 2 vecs now it's down to 1 without any noticeable change to the outer API or behaviour


Keats:
Thanks!

PR #238, opened 2022-02-09
Explained how to get JWK, edited DECODE section

rimutaka:
JWT/JWK/JWKS/OIDC terminology can be quite intimidating for the uninitiated like myself. I had to do some extra digging to figure out what JWK is, where to get it and how that machinery works. I think adding a few sentences explaining that will be helpful.

Otherwise the docs are clear and I got it working in no time. Thank you very much!

I put my edits in a separate branch in case you want to edit them further. No offense if you reject.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,14 +106,6 @@ something similar and reuse it.
 
 ### Decoding
 
-In some cases when the algorithm is not known or you need to grab the `kid`, you can choose to decode the header only:
-
-```rust
-let header = decode_header(&token)?;
-```
-
-This does not perform any signature verification or validate the token claims.
-
 #### Decoding with a shared secret
 
 ```rust
@@ -153,9 +145,17 @@ The JWK values can usually be found by looking up the OIDC configuration of the 
 2. look up the URL for `jwks_uri` field in the downloaded document
 3. download JWKS (a JWK set) from `jwks_uri` and use the record with `kid` matching `kid` in the token's header
 
-If you are the token issuer, you should be able to find the link to your JWKS somewhere in your token provider's control panel.
+Keys in PEM format have better performance if `DecodingKey` is generated once with `once_cell` for reuse.
 
-Keys in PEM format have better performance if `DecodingKey` is generated once with `lazy_static` for reuse.
+#### Decoding header only
+
+In some cases when the algorithm is not known or you need to grab the `kid`, you can choose to decode the header only:
+
+```rust
+let header = decode_header(&token)?;
+```
+
+This does not perform any signature verification or validate the token claims.
 
 ### Convert SEC1 private key to PKCS8
 `jsonwebtoken` currently only supports PKCS8 format for private EC keys. If your key has `BEGIN EC PRIVATE KEY` at the top,
--- a/README.md
+++ b/README.md
@@ -106,6 +106,16 @@ something similar and reuse it.
 
 ### Decoding
 
+In some cases when the algorithm is not known or you need to grab the `kid`, you can choose to decode the header only:
+
+```rust
+let header = decode_header(&token)?;
+```
+
+This does not perform any signature verification or validate the token claims.
+
+#### Decoding with a shared secret
+
 ```rust
 // `token` is a struct with 2 fields: `header` and `claims` where `claims` is your own struct.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,14 +147,15 @@ The main use-case is for JWK where your public key is in a JSON format like so:
...
-// `token` is a struct with 2 fields: `header` and `claims` where `claims` is your own struct.
-let token = decode::<Claims>(&token, &DecodingKey::from_rsa_components(jwk["n"], jwk["e"]), &Validation::new(Algorithm::RS256))?;
-```
+1. append `/.well-known/openid-configuration` to the value of `iss` field from the token and download the JSON file from that URL
+2. look up the URL for `jwks_uri` field in the downloaded document
+3. download JWKS (a JWK set) from `jwks_uri` and use the record with `kid` matching `kid` in the token's header
 
-If your key is in PEM format, it is better performance wise to generate the `DecodingKey` once in a `lazy_static` or
-something similar and reuse it.
+If you are the token issuer, you should be able to find the link to your JWKS somewhere in your token provider's control panel.
```
Keats, line 156: Honestly I'm not sure this part on JWKs should be in the docs. For example, it would be different if you use JWKs from Auth0/Hydra so that would be just confusing many users.

Keats (changes requested):

README.md
```diff
@@ -138,14 +139,23 @@ The main use-case is for JWK where your public key is in a JSON format like so:
    "n":"yRE6rHuNR0QbHO3H3Kt2pOKGVhQqGZXInOduQNxXzuKlvQTLUTv4l4sggh5_CYYi_cvI-SXVT9kPWSKXxJXBXd_4LkvcPuUakBoAkfh-eiFVMh2VrUyWyj3MFl0HTVF9KwRXLAcwkREiS3npThHRyIxuy0ZMeZfxVL5arMhw1SRELB8HoGfG_AtH89BIE9jDBHZ9dLelK9a184zAf8LwoPLxvJb3Il5nncqPcSfKDDodMFBIMc4lQzDKL5gvmiXLXB1AGLm8KBjfE8s3L5xqi-yUod-j8MtvIj812dkS4QMiRVN_by2h3ZY8LYVGrqZXZTcgn2ujn8uKjXLZVD5TdQ"
 }
 ```
+The JWK values can usually be found by looking up the OIDC configuration of the token issuer:
+
+1. append `/.well-known/openid-configuration` to the value of `iss` field from the token and download the JSON file from that URL
+2. look up the URL for `jwks_uri` field in the downloaded document
+3. download JWKS (a JWK set) from `jwks_uri` and use the record with `kid` matching `kid` in the token's header
```
Keats, line 146: As mentioned before I don't think I would include these in the README, it's more likely to confuse people imo. Eg I've used JWKs before but never did any of this.

rimutaka: @Keats , Vincent, sorry, I misunderstood your earlier comment! Adding this part was the main driver for the PR. It's aimed at devs who work with this kind of stuff for the first time or not often enough to remember all the ins and outs. This method is not dependent on the Auth provider. If you know an easier way of getting the keys I'd love to know it :)

There is no point applying the rest of the changes if this one is out. I'll kill the PR. Sorry for wasting your time.

Keats: So for example for my usage of JWKs, I do a HTTP request to my oauth server `.well-known/jwks.json` which gives me something like https://github.com/Keats/jsonwebtoken/blob/master/examples/auth0.rs#L8
Basically https://github.com/Keats/jsonwebtoken/blob/master/examples/auth0.rs is an example of my usage of JWKs. I've never really seen or used the steps from your PR, that's why I find it a bit confusing

rimutaka:
Ha! And that's what I assumed too. Turns out, there is no spec that says so. The JWK RFC (https://datatracker.ietf.org/doc/html/rfc7517) doesn't mention any location for JWKS at all. The OIDC spec mentions `/.well-known/openid-configuration` in https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig and it is registered with IANA (https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml).

Technically, the JWK file can be named anything and placed anywhere, but I have no example of it being anything other than `[iss]/.well-known/jwks.json`. 

P.S. not trying to argue with you - just sharing my findings :)

PR #240, opened 2022-02-20
Made spec_claims exp, nbf, iss, sub, and aud optional, unless they are specifically required

cloutiertyler:

PR #247, opened 2022-04-05
Fix #190 / cleanup of PR #240

### LINKED ISSUE #190: The exp claim should be optional

brocaar:
This PR contains the changes made by @cloutiertyler. The only thing that I have done is ran `cargo fmt` and fixed the feedback provided by `cargo clippy`. This solves issue #190.

### DIFF
--- a/src/validation.rs
+++ b/src/validation.rs
@@ -230,46 +230,49 @@ pub(crate) fn validate(claims: ClaimsForValidation, options: &Validation) -> Res
         }
     }
 
-    if options.validate_exp
-        && !matches!(claims.exp, TryParse::Parsed(exp) if exp >= now-options.leeway)
-    {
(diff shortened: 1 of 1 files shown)


cloutiertyler:
Awesome! Are there any additional tests we should have or do the tests I added look good?

Keats:
I think it's fine! If you both confirmed it works for you, I'll merge and release sometimes this week

Keats:
That should be 8.1 and not 9.0 right?

cloutiertyler:
👏 Thank you!

PR #256, opened 2022-06-09
Fix invalid field name on OctetKeyParameters

jtdavey:
RFC 7518 defines the parameters for symmetric keys to have the 'k' (Key
Value) parameter (Section 6.4.1). The OctetKeyParameters structure
unfortunately has it defined as "value".

This change should allow parsing of jwks.json files with a symmetric key
while keeping the same api.

I've also added a simple test to validate this works as expected.


jtdavey:
It looks like a maintainer needs to approve running the workflow? I didn't have CLion setup to run rustfmt on save for this one, so missed it.

PR #265, opened 2022-10-12
Fix algorithm family validation

Jonathas-Conceicao:
I've noticed that when `Validation` has algorithms of different families
only the first one was being validated.

For example. if we have:
```Rust
let mut validation = jsonwebtoken::Validation::new(jsonwebtoken::Algorithm::RS256);
validation.algorithms.push(jsonwebtoken::Algorithm::RS384);
validation.algorithms.push(jsonwebtoken::Algorithm::ES384);
validation.algorithms.push(jsonwebtoken::Algorithm::ES256);
```

Only `AlgorithmFamily::Rsa` would be accepted in the JWT instead of
`AlgorithmFamily::Ec` also being acceptable.

This change ensures that all algorithm families are checked.

### DIFF
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -157,10 +157,8 @@ fn verify_signature<'a>(
     }
 
     if validation.validate_signature {
-        for alg in &validation.algorithms {
-            if key.family != alg.family() {
-                return Err(new_error(ErrorKind::InvalidAlgorithm));
-            }
+        if !validation.algorithms.iter().any(|alg| key.family == alg.family()) {
+            return Err(new_error(ErrorKind::InvalidAlgorithm));
         }
     }
 


otavio (approved):

Keats:
I'm not very satisfied with this bit tbh, I'd want to be able to specify only from a family to avoid weird behaviours like that

otavio:
But this would require an API change, wouldn't it?

Keats:
Yep it would, I'm just thinking what it could look like.

Jonathas-Conceicao:
Yeah, I agree that the current API for adding expected algorithms is misleading. I think making the `AlgorithmFamily` enum public and using it instead would show the expected behavior more clearly.

I do, however, feel like the `Validation` should still support checking against different families, as a particular protocol I was implementing supports both RSA and EC for its JWT.

Jonathas-Conceicao:
Either way, I'm closing this, as I think the PR is pointless if the intention is to rework the public API.

Keats:
> I do, however, feel like the Validation should still support checking against different families, as a particular protocol I was implementing supports both RSA and EC for its JWT.

I think it's better to check the header to see the alg rather than letting the library pick the alg among multiple families. We don't have the None alg so that's better but still

PR #267, opened 2022-10-23
Creating DecodingKey from jwk

cduvray:
- adding DecodingKey:from_ec_components DecodingKey:from_ed_components (equivalent of existing DecodingKey:from_rsa_components for ECDSA and EdDSA)
- adding tests
- adding DecodingKey:from_jwk
- adding tests for rsa, ecdsa and eddsa

### DIFF
--- a/tests/rsa/mod.rs
+++ b/tests/rsa/mod.rs
@@ -127,6 +127,41 @@ fn rsa_modulus_exponent() {
     assert!(res.is_ok());
 }
 
+#[cfg(feature = "use_pem")]
+#[test]
+fn rsa_jwk() {
+    use jsonwebtoken::jwk::Jwk;
+    use serde_json::json;
+
+    let privkey = include_str!("private_rsa_key_pkcs8.pem");
+    let my_claims = Claims {
+        sub: "b@b.com".to_string(),
+        company: "ACME".to_string(),
+        exp: OffsetDateTime::now_utc().unix_timestamp() + 10000,
+    };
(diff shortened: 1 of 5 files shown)


tests/rsa/mod.rs
```diff
@@ -127,6 +127,41 @@ fn rsa_modulus_exponent() {
...
+        &EncodingKey::from_rsa_pem(privkey.as_ref()).unwrap(),
+    )
+    .unwrap();
+    let res = decode::<Claims>(
+        &encrypted,
+        &DecodingKey::from_jwk(&jwk).unwrap(),
+        &Validation::new(Algorithm::RS256),
+    );
+    assert!(res.is_ok());
+}
```
Keats, line 163: should those tests use decode_from_jwk?

cduvray: I'm not sure I understand well ..

the tests/rsa/rsa_jwt() is  a round trip: encoding -> decoding using `DecodingKey::from_jwk()` to build the decoding key. (this is an equivalent of  the test `rsa_modulus_exponent()`, I used that one as a "model" for my tests)

There are also tests for `DecodingKey::from_ed_components` and `DecodingKey::from_ec_components`, The implementation of DecodingKey::from_jwk() is based on these 2 functions. I made them public because they are "counterparts" of `from_rsa_components()`  for ECDSA et EdDSA keys.

cduvray: Hello @Keats any precisions on how I should modify tests? or other suggestions?

PR #274, opened 2022-12-01
CERTIFICATE tag for pem files

Alexey-N-Chernyshov:
Fix for issues #259, surrealdb/surrealdb#1256.

Adds `CERTIFICATE` tag for pem files.
RSA tests were extended with certificates.
Self-signed certificates were generated as follows:
```
# openssl req -new -key private.pem -out certificate.csr
# openssl x509 -req -days 358000 -in certificate.csr -signkey private.pem -out certificate.crt
```
certificate.crt files may be checked at https://www.sslchecker.com/certdecoder

Also there are some fixes found by clippy in the first two commits.


Keats:
Can you add a note somewhere in the test folder next to the certificate for the commands? I know it's in git blame but it's just that I'd like to put the other commands eventually as I never remember them

src/pem/decoder.rs
```diff
@@ -79,41 +79,43 @@ impl PemEncodedKey {
                     // "there is no such thing as a "PKCS#1 format" for elliptic curve (EC) keys"
 
                     // This handles PKCS#8 public & private keys
```
Keats, line 81: This comments needs to be updated

Keats (approved):
LGTM, just some minor comments.

src/pem/decoder.rs
Alexey-N-Chernyshov: Done

Alexey-N-Chernyshov:
Also I added a note with commands in tests/rsa/README.md.

Can you make a release, please?

Keats (approved):

PR #276, opened 2022-12-07
feature (ci): dependabot for cargo and github-actions

Alexey-N-Chernyshov:
Reopened #275 on updated branch.


Keats:
Thanks for the PR. I'm not a fan of dependabot though, it tends to create so much noise and I'm only using it for a few repos :/
I usually update dependencies on new release and if a new dep need to be updated, people will do an issue/PR pretty fast usually.

PR #282, opened 2023-01-04
added an example for ed25519

ovidiu-ionescu:
Added an example for ed25519 as using it is not obvious from the documentation.


Keats (approved):

Keats:
Thanks!

PR #283, opened 2023-01-07
style: use non-reference patterns instead of explicit ref in patterns

samueltardieu:
By using non-reference patterns to match references, the `ref` keyword becomes unnecessary as it is added by the desugaring.

### DIFF
--- a/examples/auth0.rs
+++ b/examples/auth0.rs
@@ -18,8 +18,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         None => return Err("Token doesn't have a `kid` header field".into()),
     };
     if let Some(j) = jwks.find(&kid) {
-        match j.algorithm {
-            AlgorithmParameters::RSA(ref rsa) => {
+        match &j.algorithm {
+            AlgorithmParameters::RSA(rsa) => {
                 let decoding_key = DecodingKey::from_rsa_components(&rsa.n, &rsa.e).unwrap();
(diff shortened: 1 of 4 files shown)


Keats:
I'm still on the fence about this elision on whether it's good or not

samueltardieu:
I've pushed a reformatted version so that `cargo format` doesn't complain.

> I'm still on the fence about this elision on whether it's good or not

Non-reference patterns have been introduced in Rust 1.26 in an effort to make Rust `match` [more ergonomic](https://github.com/rust-lang/rfcs/blob/master/text/2005-match-ergonomics.md) and avoid the explicit use of `ref`.

samueltardieu:
Also note that non-reference patterns are already used in the code, for example at `src/decoding.rs:180` (to make `b` a `&[u8]`):

```rust
    pub(crate) fn as_bytes(&self) -> &[u8] {
        match &self.kind {
            DecodingKeyKind::SecretOrDer(b) => b,
            DecodingKeyKind::RsaModulusExponent { .. } => unreachable!(),
        }
    }
```

or in `src/crypto/mod.rs:99` (to make `bytes` a `&[u8]`):

```rust
            match &key.kind {
                DecodingKeyKind::SecretOrDer(bytes) => verify_ring(alg, signature, message, bytes),
                DecodingKeyKind::RsaModulusExponent { n, e } => {
                    rsa::verify_from_components(alg, signature, message, (n, e))
                }
            }
```

PR #291, opened 2023-01-30
Update actions/checkout version

giraffate:
`actions/checkout@v3` is available now: https://github.com/actions/checkout/blob/main/CHANGELOG.md

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     name: Format
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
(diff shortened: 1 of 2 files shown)


giraffate:
The Clippy test fails because the new lint was added in 1.67. I'll prepare a fix for it tomorrow.

Keats:
Yeah I just got the same on another repo. I'm tempted to remove ` -D warnings` since i don't want some clippy warnings are kind of opinionated and don't have a notion of MSRV

giraffate:
Clippy can support the MSRV by setting the msrv tag in clippy.toml. Does this project have that version? I can't find it.

giraffate:
> Clippy can support the MSRV by setting the msrv tag in clippy.toml. Does this project have that version? I can't find it.

I missed it in README. And I set `rust-version` in Cargo.toml instead of adding clippy.toml.

Keats (approved):

PR #298, opened 2023-03-13
Implement Clone for TokenData<T> if T: Clone

fiadliel:
If T: Clone, we can write a Clone implementation for `TokenData`.


Keats (approved):

fiadliel:
@Keats Thanks for the review! Just a small note, I can't merge 😄

PR #302, opened 2023-03-22
fix(decoding): do not get_current_timestamp if user dont validate exp/nbf claims

Oliboy50:
## Why?
Because getting current timestamp has a (small) cost that we can avoid when user don't ask for validating "exp"/"nbf" claims.


Keats:
Thanks!

PR #303, opened 2023-03-24
Update pem dependency to 2.x

urkle:
- updating this removes the duplicate base64 dependency as pem 1.x used base64 0.13 still

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ serde = {version = "1.0", features = ["derive"] }
 ring = { version = "0.16.5", features = ["std"] }
 base64 = "0.21.0"
 # For PEM decoding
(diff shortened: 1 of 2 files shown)


giraffate:
The pem 2.0 requires 1.67 for MSRV. So, pinned tests in CI fails because it's 1.59.

Keats:
1.59 -> 1.67 is a pretty steep change!

urkle:
yeah.. that is a big jump I did not realize that pem jumped that high.   Sucks as that's the only way to not have the duplicate base64 crate. 

So perhaps this should be postponed until you are ready to jump MSRV

Keats:
It's been 6 months now, maybe it's time to do the bump

PR #307, opened 2023-04-10
Update actions to avoid Node.js 12 actions

giraffate:
Suppress these warnings: https://github.com/Keats/jsonwebtoken/actions/runs/4430067735


Keats:
Thanks. It looks like https://github.com/dtolnay/rust-toolchain is the most common replacement looking at https://github.com/actions-rs/toolchain/issues/216

Issue #300, opened 2023-03-22
Expiration validated incorrectly when system time is not UTC


litcc:
Related questions: https://github.com/Keats/jsonwebtoken/issues/253#issue-1253005342

https://github.com/Keats/jsonwebtoken/blob/8fba79b25459eacc33a80e1ee37ff8eba64079ca/src/validation.rs#L12-L38

https://github.com/Keats/jsonwebtoken/blob/8fba79b25459eacc33a80e1ee37ff8eba64079ca/src/validation.rs#L214-L231

https://github.com/Keats/jsonwebtoken/blob/8fba79b25459eacc33a80e1ee37ff8eba64079ca/src/validation.rs#L138-L144

The comment of Validation states that all validation is based on UTC timestamps, but in the actual code, the time obtained is the current time in the current system time zone, so there may be ambiguities, and when exp is passed in UTC timestamps, the time validation will not take effect.

Keats:
Hmm I guess everyone runs their server with UTC. We could pull chrono (or any other lib giving the local offset) as a feature to determine the offset and then convert it to UTC timestamp

litcc:
then the documentation can be amended to state that it does not necessarily complete all time checks via UTC, depending on the current system time zone.

It is also possible to add some configuration to Validationt, such as providing a user-configurable source for the current time, setting fn get_current_timestamp() -> u64 as the default source, which would avoid relying directly on third-party time libraries in the library and let the user choose which one to use

qsantos:
Should this bug be closed? According to @nickguletskii in #311, this does not actually happen. If this is a real bug, it's critical, since this directly affect security and can be overlooked extremely easily.

I did encounter an issue with expiration validation, which led me here, but I just did not know about [`Validation::leeway`](https://docs.rs/jsonwebtoken/latest/jsonwebtoken/struct.Validation.html#structfield.leeway). When taking it into account, I do not have any issue with expiration validation even though my system is currently UTC+0200.

litcc:
Previously occurred out of time zone problems, unified adjustment of the server to use the UTC time zone after the repair, currently tested by the inability to reproduce the problem, so closed

qsantos:
Thanks!

PR #311, opened 2023-05-19
Add optional support for 3rd party time libraries like `time` and `chrono`

nickguletskii:
Should resolve timezone-related issues mentioned in #300 .

This PR introduces 4 new cargo features:

* `time`, which adds optional support for deserializing and comparing timestamps using `time` using `validate_with_options`.
* `chrono`, which adds optional support for deserializing and comparing timestamps using `chrono` using `validate_with_options`. 
* `default_time`, which makes `time` the default library used by `validate` to validate the tokens,
* `default_chrono`, which makes `chrono` the default library used by `chrono` to validate the tokens,

Therefore, the changes should, at least in theory, be backwards compatible, and fixing the time zone issue should be as simple as adding `default_time` or `default_chrono` to the list of cargo features.

However, this PR does make the `custom_time` example obsolete and it will have to be rewritten at some point. It still works fine from what I can tell, it's just not necessary because using a different type for timestamps is now as easy as copying and adapting the implementations in `jsonwebtoken::time`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,3 +60,9 @@ jobs:
 
     - name: Run tests no features
       run: cargo test --no-default-features
+
+    - name: Run tests with time
+      run: cargo test --features default_time
+
+    - name: Run tests with chrono
+      run: cargo test --features default_chrono
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,9 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
-      - run: cargo clippy --all-targets --all-features -- -D warnings
+      - run: cargo clippy --all-targets --features pem -- -D warnings
+      - run: cargo clippy --all-targets --features pem,default_time -- -D warnings
+      - run: cargo clippy --all-targets --features pem,default_chrono -- -D warnings
 
   tests:
     name: Tests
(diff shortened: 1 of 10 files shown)


Keats:
Thanks but I'm not sure it helps?

I'm pretty sure `time` and `chrono` will just call out `SystemTime` anyway for UTC (eg https://docs.rs/chrono/latest/src/chrono/offset/utc.rs.html#67-73 and  https://github.com/time-rs/time/blob/main/time/src/date_time.rs#L323) so that doesn't change the issue?

I'm in UTC+2 and I do get the same timestamp with time and SystemTime as expected. AFAIK the only way to solve that get the timezone offset of the server as part of the validation.

You can have only the `time` optional dep (no need for chrono) and in `get_current_timestamp` in a cfg if time dep enabled grab the current local offset (https://time-rs.github.io/api/time/struct.UtcOffset.html#method.current_local_offset) and apply that to the timestamp. No need for other changes I think.

nickguletskii:
Sorry, you are right. I think the confusion comes from how things are named in `std::time`;

The issuer of the token must always use UTC for expiry times. That much is clear and the author of the original issue seems to agree with this.

What I think the original issue is actually claiming is that the UTC timestamps in the tokens are being compared to local time. However, that is not the case, because `std::time::SystemTime` is not the system's local time (i.e. time with local timezone), it's the system's UTC clock. Therefore, the timestamp validation in jsonwebtoken is correct and there's no need for this PR.

PR #313, opened 2023-06-05
Upgrade pem from version 1 to 2

vojd:
Upgrading `pem` crate from version `1` to `2` in order to get the latest `base64` crate version `0.21`


vojd:
Thank you!

Issue #314, opened 2023-06-07
The pinned tests fails


giraffate:
https://github.com/Keats/jsonwebtoken/actions/runs/5027342878/jobs/9016716260#step:5:77
> error: package `regex v1.8.1` cannot be built because it requires rustc 1.60.0 or newer, while the currently active rustc version is 1.59.0

The pinned version is 1.59.0. The regex is a part of the dev-dependencies. So, would the pinned version be updated to 1.60.0? (In this project, I don't know that whether the pinned version is different from msrv.)

Keats:
Hmm that's an interesting situation. I guess we can pin criterion for now since it's dev only? (I'm guessing regex is in criterion)

giraffate:
Yes, regex is dependencies of criterion. I don't know this is the good way, but it works when we pin regex directly in dev-dependencies.
https://github.com/Keats/jsonwebtoken/pull/315

Keats:
Fixing it by bumping the pinned version. Thanks!

PR #315, opened 2023-06-08
Pin regex version in dev-dependencies

giraffate:
Fix https://github.com/Keats/jsonwebtoken/issues/314

The regex v1.8.x needs the rustc 1.60.0 or newer, so we pins its version.


giraffate:
Oh, Clippy in CI fails, I'll address it later.

Keats:
Maybe it's ok to do https://github.com/Keats/jsonwebtoken/pull/303 now that it's been 6 months since 1.67 release

giraffate:
Okay, it looks good!

PR #316, opened 2023-06-12
JWK hmac keys are base64url encoded not standard

sjoerdsimons:
DecodingKey::from_base64_secret uses standard decoding, while "k" field of a jwk is expected to be in base64url instead.


Keats:
Can you add a test that was failing before?

sjoerdsimons:
> Can you add a test that was failing before?

Done; without the first commit this test would fail with:
`thread 'verify_hs256_rfc7517_appendix_a1' panicked at 'called `Result::unwrap()` on an `Err` value: Error(Base64(InvalidByte(34, 45)))', tests/hmac.rs:199:43`

PR #320, opened 2023-06-21
Allow validating against user provided timestamp

jakubtrnka:


src/validation.rs
```diff
@@ -25,6 +25,8 @@ use crate::errors::{new_error, ErrorKind, Result};
 /// // or issuer
 /// validation.set_issuer(&["Me"]); // a single string
 /// validation.set_issuer(&["Me", "You"]); // array of strings
+/// validation.validate_against_system_time(Some(1687425589)) // validate expiry and
```
Keats, line 28: That's a niche function, let's not add it to the doc of the struct that would be confusing

src/validation.rs
```diff
@@ -98,9 +104,16 @@ impl Validation {
             aud: None,
 
             validate_signature: true,
+            timestamp: None,
         }
     }
 
+    /// If set with Some, the value will be used to validate expiry and not-valid-after fields.
+    /// If set to None, local SystemTime will be used
+    pub fn validate_against_system_time(&mut self, timestamp: Option<u64>) {
```
Keats, line 113: I think `fn use_given_timestamp(&mut self, timestamp: u64)` is a better signature. You're never going to call that with `None`.

tests/ecdsa/mod.rs
```diff
@@ -72,6 +74,46 @@ fn round_trip_claim() {
     assert_eq!(my_claims, token_data.claims);
 }
 
+#[cfg(feature = "use_pem")]
+#[test]
+fn expiry_and_nbf() {
```
Keats, line 79: make it explicit in the name it's testing the new feature

Keats (changes requested):

src/validation.rs
jakubtrnka: Ok. I'll change the signature.

Now that I'm thinking about this function more, I think the implementation should actually be following
```rs
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct Validation {
    // ...
    /// The validation will check that this value lies between expiry and not-valid-before
    /// timestamps
    pub(crate) timestamp: Option<u64, Instant>,
}

impl Validation {
    /// If set with Some, the value will be used to validate expiry and not-valid-after fields.
    /// If set to None, local SystemTime will be used
    pub fn use_given_timestamp(&mut self, timestamp: u64) {
        self.timestamp = Some((timestamp, Instant::now()));
    }

    pub fn get_current_timestamp(&self) -> u64 {
        self.timestamp.1.elapsed().as_secs() + self.0.timestamp
    }
}
```

Such that when configured with "current" timestamp. It mimics the default behavior with custom timestamp unset.
Pinning the timestamp as a fixed number in Validation struct is probably anti-feature. What do you think?

(btw in the meantime I learned I can switch off the time validation in this crate, so I did my own validation in my project and don't need this feature anymore. I can finish this since I started it, but if you don't want that feature I'm prefectly fine if you close it).

tests/ecdsa/mod.rs
jakubtrnka: I'm not sure what you mean by this.

src/validation.rs
Keats: > (btw in the meantime I learned I can switch off the time validation in this crate, so I did my own validation in my project and don't need this feature anymore. I can finish this since I started it, but if you don't want that feature I'm prefectly fine if you close it).

I think I'll close it then, better to let edge cases be handled outside of the library. Thanks for the work though!

Issue #252, opened 2022-05-24
JWK `alg` types are missing and deserialization from spec-compliant JWKS will fail


jblachly:
Hi,

First, thanks for this library which has been really helpful and is built in much the same way I would have built my own JWT/JWK library.

# Background

When building an API that works with Keycloak and reads its JWKS (JSON web key set), I found that I could not properly deserialize the JWKS and my program panicked on startup. **Ultimately, this is because one of the JWK has an `alg` field that is not a member of the `Algorithm` enum[1].**  However, this JWK is allowed by the specs (details below) and thus we cannot deserialize a spec-compliant JWK. 

# Detailed problem

JWK specifications [2] allow an `alg` field to take the same values as allowed in JWS (JSON web signature; these values match the `Algorithm` enum variants and include things like `HS256`, `RS256`, `ES256`, etc.; see ref [3]) , but _also_ to take JWE (JSON web encryption) alg types defined in [4]. These include principally `RSA-OAEP` and `ECDH-ES`. Note that these values would ALSO be valid values for the `alg` parameter of a JWT if the JWT were encrypted (and not just signed).

I can write a patch and PR for this, but I wanted to get an ACK or NACK on the overall idea, and also on specific fix that would be backward-compatible.

# Proposal for backward-compatible fix

To minimize complexity, admix the `alg` types from JWS [3] and JWE [4] as variants of the `Algorithm` enum [1]. This would be an easy fix.

A more complex fix might involve separating the algorithms into signing algorithms and encryption algorithms, but I don't see a ton of value in this unless additional API work is done to the crate to introspect whether a key was simply signed, or whether was encrypted. This would likely be backward-incompatible/API breaking.

@Keats let me know what you think and I'll put it together ASAP

# Example JWKS with two JWK keys:
```
[...]

PR #324, opened 2023-08-16
deserialization support for some missing algorithms

nick9822:
Fix towards #252 
Added deserialization support for some missing algorithms in order to avoid panic while deserialization.

### DIFF
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -1,6 +1,7 @@
 use crate::errors::{Error, ErrorKind, Result};
+use ring::hmac::Key;
 use serde::{Deserialize, Serialize};
-use std::str::FromStr;
+use std::{fmt, str::FromStr};
 
 #[derive(Debug, Eq, PartialEq, Copy, Clone, Serialize, Deserialize)]
 pub(crate) enum AlgorithmFamily {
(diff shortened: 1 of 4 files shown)


src/algorithms.rs
```diff
@@ -80,6 +81,86 @@ impl Algorithm {
             Algorithm::EdDSA => AlgorithmFamily::Ed,
         }
     }
+
+    /// Converting Key Algorithm to Algorithm
+    pub fn frm_key_alogorithm(s: &KeyAlgorithm) -> Result<Self> {
```
schitcrafter, line 86: Thanks for the pr, I couldn't figure out how to best implement this myself, but - I think this should spell "from_key_algorithm"? And maybe implementing From<KeyAlgorithm> would be nice too, but I'm not sure

src/jwk.rs
```diff
@@ -326,6 +327,16 @@ pub struct Jwk {
     pub algorithm: AlgorithmParameters,
 }
 
+impl Jwk {
+    /// Find whether the Algorithm is implmented and supported
+    pub fn is_supported(&self) -> bool {
+        match Algorithm::frm_key_alogorithm(&self.common.key_algorithm.unwrap()) {
```
Keats, line 333: `Algorithm::from_key_alogorithm(&self.common.key_algorithm.unwrap()).is_ok()`

src/algorithms.rs
```diff
@@ -80,6 +81,86 @@ impl Algorithm {
...
+        Algorithm::from_str(s.to_string().as_str())
+    }
+}
+
+/// The algorithms of the keys
+#[allow(clippy::upper_case_acronyms)]
+#[derive(Debug, PartialEq, Eq, Hash, Copy, Clone, Serialize, Deserialize)]
+pub enum KeyAlgorithm {
```
Keats, line 94: should this leave in the jwk file since it's only used there?

Keats (changes requested):

src/algorithms.rs
nick9822: Thanks for the comment. From trait signature would not match with us, I think From makes sense when we expect implicit conversion which will work almost always. Here the conversion is explicit and result in an error.

Btw renaming `frm_key_alogorithm` to `from_key_alogorithm`.

src/algorithms.rs
nick9822: But isn't it better be inside `algorithms` file? That will keep both enums visually in check.

nick9822:
@Keats `suspicious_doc_comments`

```
error: this is an outer doc comment and does not apply to the parent module or crate
 --> src/jwk.rs:2:1
```
Let me know if you want me to remove `!` from the comments.

src/algorithms.rs
Keats: Not really I think. I'd rather have a kinda self-contained jwk.rs rather than having to see definitions in various places

src/jwk.rs
```diff
@@ -326,6 +327,13 @@ pub struct Jwk {
     pub algorithm: AlgorithmParameters,
 }
 
+impl Jwk {
+    /// Find whether the Algorithm is implmented and supported
+    pub fn is_supported(&self) -> bool {
+        Algorithm::from_key_alogorithm(&self.common.key_algorithm.unwrap()).is_ok()
```
Keats, line 333: still a typo on the fn name

src/algorithms.rs
nick9822: Sure. Let's move it.

src/algorithms.rs
```diff
@@ -80,6 +83,11 @@ impl Algorithm {
             Algorithm::EdDSA => AlgorithmFamily::Ed,
         }
     }
+
+    /// Converting Key Algorithm to Algorithm
+    pub fn from_key_algorithm(s: &KeyAlgorithm) -> Result<Self> {
```
Keats, line 88: I think it makes more sense to do `KeyAlgorithm.to_algorithm` since this is not really meant to be public in practice and this way all the logic is contained in jwk.rs?

nick9822: Agreed. I previously thought so but in the example `auth0.rs` it needed `Algorithm` from `KeyAlgorithm` at `Validation::new`. I think we don't need public method just for that, we can do it via `Algorithm::from_str`

src/algorithms.rs
```diff
@@ -99,4 +101,19 @@ mod tests {
...
+            "PS512", "RS512", "EdDSA",
+        ];
+
+        for s in supported_algs {
+            assert!(
+                Algorithm::from_str(s).unwrap() as isize
+                    == KeyAlgorithm::from_str(s).unwrap() as isize
+            );
+        }
+    }
```
Keats, line 118: let's remove that test, I don't want tests to rely on enum ordering

Keats (approved):

schitcrafter:
Any update on this? It's been approved for some time now, but still hasn't gotten merged, would be nice to atleast know the reason why

Keats:
I forgot about it

Issue #220, opened 2021-11-17
`Debug` and `PartialEq` should not be derived for `EncodingKey` and `DecodingKey`


LukasKalbertodt:
Both of these types potentially contain secrets in a `Vec<u8>`. 

Deriving `Debug` means the secret is printed when using that impl. It can easily happen that this type is in some other struct, which is contained in some other struct, which is then logged accidentally. And suddenly you have leaked secrets in your log file. I would recommend manually deriving `Debug` and not including the actual secret value. For exampled `secrecy::Secret` prints `REDACTED` instead.

The derived `PartialEq` will result in a non-constant-time algorithm. While unlikely, if the key is ever compared to some value that an attacker can supply, then the attacker can employ a timing attack to extract the secret. I am not sure if the `PartialEq` impl is ever used. If not, I would recommend removing it. Otherwise, the derived impl should be replaced with a constant-time algorithm.

Keats:
Good catch! I'll remove them in the `next` branch.

PR #328, opened 2023-09-24
Implement Eq + Hash on `DecodingKey`

baloo:


Keats:
See https://github.com/Keats/jsonwebtoken/issues/220

baloo:
I'd agree with `Debug` here, but `Hash` and `Eq` are useful when putting a set in an evmap (https://docs.rs/evmap/10.0.2/evmap/struct.WriteHandle.html). That was my usecase.

I can't even argue `DecodingKey` only holds public keys as it's also used with Hmac based schemes with shared secrets.
I guess I'll need to wrap my keys and implement that in a wrapper of some kind.

PR #332, opened 2023-10-08
Reject tokens when claims has an aud, none expected

### LINKED ISSUE #329: jsonwebtoken::Validation could help users implement the spec more correctly
RFC7519 has this to say about the `aud` field:

> Each principal intended to process the JWT MUST
> identify itself with a value in the audience claim.  If the principal
> processing the claim does not identify itself with a value in the
> "aud" claim when this claim is present, then the JWT MUST be
> rejected.

I'm sure many users aren't using this field, and are opening themselves up to vulnerabilities of jwts intended for one use and reused for another. What if the API was adjusted to be something like this, to help users figure out they should be paying attention here?

```
Validation::new(Algorithm::HS256, Some(&["me"]);
```

This is different from the other fields which are optionally set, since those are generally configured to be relatively secure out of the box.

grahamc:
From the RFC:

> Each principal intended to process the JWT MUST
> identify itself with a value in the audience claim. If the principal
> processing the claim does not identify itself with a value in the
> "aud" claim when this claim is present, then the JWT MUST be
>rejected.

Closes #329

### DIFF
--- a/src/validation.rs
+++ b/src/validation.rs
@@ -263,6 +263,11 @@ pub(crate) fn validate(claims: ClaimsForValidation, options: &Validation) -> Res
     }
 
     match (claims.aud, options.aud.as_ref()) {
+        // Each principal intended to process the JWT MUST
+        // identify itself with a value in the audience claim. If the principal
+        // processing the claim does not identify itself with a value in the
+        // "aud" claim when this claim is present, then the JWT MUST be
+        //  rejected.
         (TryParse::Parsed(_), None) => {
             return Err(new_error(ErrorKind::InvalidAudience));
         }
--- a/src/validation.rs
+++ b/src/validation.rs
@@ -263,6 +263,9 @@ pub(crate) fn validate(claims: ClaimsForValidation, options: &Validation) -> Res
     }
 
     match (claims.aud, options.aud.as_ref()) {
+        (TryParse::Parsed(_), None) => {
+            return Err(new_error(ErrorKind::InvalidAudience));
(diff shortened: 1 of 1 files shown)


src/validation.rs
```diff
@@ -263,6 +263,9 @@ pub(crate) fn validate(claims: ClaimsForValidation, options: &Validation) -> Res
     }
 
     match (claims.aud, options.aud.as_ref()) {
+        (TryParse::Parsed(_), None) => {
```
Keats, line 266: Can you put the snippet of the RFC as a comment above here as well?

grahamc:
```suggestion
        // Each principal intended to process the JWT MUST
        // identify itself with a value in the audience claim. If the principal
        // processing the claim does not identify itself with a value in the
        // "aud" claim when this claim is present, then the JWT MUST be
        //  rejected.
        (TryParse::Parsed(_), None) => {
```

Like this?

Keats: Yep

grahamc: Done :)

Keats:
Thanks. That's a breaking change so that would be released as v9

grahamc:
Thanks, @Keats! Do you have a timeline or process for these sorts of things, or is it something that "just happens" from time to time? Just curious to set my own expectations.

Keats:
Probably this week if i don't forget about it in the meantime

grahamc:
Thank you greatly!

PR #336, opened 2023-10-25
Add the ability to decode a JWT token without specifying an audience.

### LINKED ISSUE #335: `aud` mismatch rejections are hard to debug
The change made in #332 has makes it hard to figure out what is going wrong when a token is rejected for having the wrong audience. I already check for this in my application code and provide a meaningful error, but without some way to decode the JWT to understand _what `aud` was sent_ I cannot convert my code to utilize this new version without doing manual `base64` + `serde_json` deserialization.

While I agree this change is necessary to match the spec, I'd like to suggest that a similar method to `Validation::insecure_disable_signature_validation` is introduced to aid in inspecting invalid JWTs.

Storing the rejected `aud` in the error type would also work, but I suspect this is less desirable for this crate.

sagunb:
Adding this allows us to continue using this library to decode a JWT token without a secret (the way it used to be possible pre v9).

Without this we cannot update to v9 and we are stuck in v8.3.

closes #335

### DIFF
--- a/src/validation.rs
+++ b/src/validation.rs
@@ -52,7 +52,13 @@ pub struct Validation {
     ///
     /// Defaults to `false`.
     pub validate_nbf: bool,
-    /// If it contains a value, the validation will check that the `aud` field is a member of the
+    /// Whether to validate the `aud` field.
+    ///
+    /// It will return an error if the `aud` field is not a member of the audience provided.
+    ///
+    /// Defaults to `true`. Very insecure to turn this off. Only do this if you know what you are doing.
+    pub validate_aud: bool,
+    /// Validation will check that the `aud` field is a member of the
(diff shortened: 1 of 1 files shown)


sagunb:
This would also help alleviate the issue raised in https://github.com/Keats/jsonwebtoken/issues/335

Keats (approved):

sagunb:
Can you merge the PR, I can't do it because I don't have write access to the repo.

Keats:
I'll merge it soon, just waiting a bit to maybe group it with some wasm stuff

horacimacias:
I'm also interested in this. If there is anything I can help to make this PR merged please let me know.

horacimacias:
thanks!

PR #340, opened 2023-10-27
Support custom timestamp

### LINKED ISSUE #339: `decode` panics on `wasm32-unknown-unknown` when `validate_exp` is `true`
`fn get_current_timestamp() -> u64` uses `SystemTime::now()` to get the current unix timestamp. Unfortunately, `SystemTime` functions panic on `wasm32-unknown-unknown`.

There are a couple of approaches to fix this:

1. Switch to `chrono` for getting the timestamp
    ```rust
    pub fn get_current_timestamp() -> u64 {
        chrono::Utc::now().timestamp() as u64
    }
    ```
1. In `Validation`, allow the user to supply a timestamp
    ```rust
    /// Supply a custom timestamp for checking `exp` and `nbf` claims.
    /// If `None`, will use `SystemTime::now()`.
    ///
    /// Defaults to `None`.
    pub current_timestamp: Option<u64>,
    ```

As a long-term solution, I prefer option 2 as it adds no cost, doesn't inflate the dependency tree, and allows for better customization.

BrandonDyer64:
This will grant the ability to use `jsonwebtoken`'s `decode` function on `wasm32-unknown-unknown` so long as a timestamp is provided. This may also help with future unit testing.

resolves #339

### DIFF
--- a/src/validation.rs
+++ b/src/validation.rs
@@ -52,6 +52,11 @@ pub struct Validation {
     ///
     /// Defaults to `false`.
     pub validate_nbf: bool,
+    /// Supply a custom timestamp for checking `exp` and `nbf` claims.
+    /// If `None`, will use `SystemTime::now()`.
+    ///
+    /// Defaults to `None`.
+    pub current_timestamp: Option<u64>,
     /// If it contains a value, the validation will check that the `aud` field is a member of the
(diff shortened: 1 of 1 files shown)


Keats:
See my comment in the PR, I don't think this is a good solution.

PR #341, opened 2023-10-29
Do what chrono does

### LINKED ISSUE #339: `decode` panics on `wasm32-unknown-unknown` when `validate_exp` is `true`

BrandonDyer64:
resolves #339 

I really don't like this approach though.

Why is setting the expiration time user controlled, but not validation? I have to import `chrono` so I can specify a timestamp for `exp`, but then the validation call is a black box that does something outside of user control. IMO, there shouldn't be a `get_current_timestamp` call. The user should have to supply one on calling `decode`, the same way they do when calling `encode`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,12 +15,18 @@ rust-version = "1.67.0"
 [dependencies]
 serde_json = "1.0"
 serde = {version = "1.0", features = ["derive"] }
-ring = { version = "0.17.4", features = ["std", "wasm32_unknown_unknown_js"] }
 base64 = "0.21.0"
 # For PEM decoding
 pem = {version = "3", optional = true}
 simple_asn1 = {version = "0.6", optional = true}
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+ring = { version = "0.17.4", features = ["std"] }
+
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
 js-sys = "0.3"
+ring = { version = "0.17.4", features = ["std", "wasm32_unknown_unknown_js"] }
 
 [dev-dependencies]
 # For the custom time example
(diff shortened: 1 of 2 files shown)


Keats:
> IMO, there shouldn't be a get_current_timestamp call. The user should have to supply one on calling decode, the same way they do when calling encode

That would be horrendous UX. Is there a single JWT library in any languages doing that by default? I can potentially image having it to override the current timestamp for testing purposes but it's present in many languages. I've checked node, python, ruby and .net and only node allows setting it for decode.

Getting the current timestamp has nothing to do with validation

Cargo.toml
```diff
@@ -15,11 +15,12 @@ rust-version = "1.67.0"
 [dependencies]
 serde_json = "1.0"
 serde = {version = "1.0", features = ["derive"] }
-ring = { version = "0.17.4", features = ["std"] }
+ring = { version = "0.17.4", features = ["std", "wasm32_unknown_unknown_js"] }
 base64 = "0.21.0"
 # For PEM decoding
 pem = {version = "3", optional = true}
 simple_asn1 = {version = "0.6", optional = true}
+js-sys = "0.3"
```
Keats, line 23: This and the features need to be optional

Keats (approved):

PR #346, opened 2023-11-19
Add wasm tests using was-bindgen-test and wasm-pack

lpotthast:
A few notes:

1) Wasm32 related tests cannot be run that easily, see https://github.com/rustwasm/team/issues/173. You have to use https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/usage.html and `wasm-pack`.

2) The `criterion` dev-dependency, only used for benchmarks, prohibits any compilation to the wasm32 target when using its default features (rayon in particular). This is now dependant on the targeted architecture.

3) Examples could not be compiled to wasm32, or any 32 bit target. The `Claims` struct from both `custom_hader.rs` and `validation.rs` had a `usize` field which got initialized with a value outside the u32 (usize on wasm32) range. Switching to an explicit u64 type in these structs seemed to be a reasonable fix?

4) Most crates provide some `wasm-...` feature, to enable wasm support. This crate uses `[target.'cfg(target_arch = "wasm32")'.dependencies]` inside its Cargo.toml. This is obviously easier for users, as there is no feature which could be forgotten. Are there any downsides to this method?

5) CI will most likely still not work, as 
  - nodejs must be installed
	https://github.com/nodesource/distributions/blob/master/README.md#ubuntu-versions
  - wasm-pack must be installed
	`cargo install wasm-pack`

Example `wasm-pack` output:
```
lukas@[...]:~/dev/jsonwebtoken$ wasm-pack test --node
[INFO]: 🎯  Checking for the Wasm target...
   Compiling jsonwebtoken v9.1.0 (/home/lukas/dev/jsonwebtoken)
    Finished dev [unoptimized + debuginfo] target(s) in 0.84s
   Compiling jsonwebtoken v9.1.0 (/home/lukas/dev/jsonwebtoken)
    Finished test [unoptimized + debuginfo] target(s) in 0.28s
     Running unittests src/lib.rs (target/wasm32-unknown-unknown/debug/deps/jsonwebtoken-ae8fe4284f1b721f.wasm)
no tests to run!
     Running tests/hmac.rs (target/wasm32-unknown-unknown/debug/deps/hmac-14855738297bd30c.wasm)
no tests to run!
     Running tests/lib.rs (target/wasm32-unknown-unknown/debug/deps/lib-17a50cd144c5e104.wasm)
Set timeout to 20 seconds...
running 21 tests                                  

test lib::ecdsa::roundtrip_with_jwtio_example ... ok
test lib::ecdsa::ed_jwk ... ok
test lib::ecdsa::ec_x_y ... ok
test lib::ecdsa::round_trip_claim ... ok
test lib::ecdsa::round_trip_sign_verification_pem ... ok
test lib::ecdsa::round_trip_sign_verification_pk8 ... ok
test lib::rsa::roundtrip_with_jwtio_example_jey ... ok
test lib::rsa::rsa_jwk ... ok
test lib::rsa::rsa_modulus_exponent ... ok
test lib::rsa::round_trip_claim ... ok
test lib::rsa::round_trip_sign_verification_der ... ok
test lib::rsa::round_trip_sign_verification_pem_pkcs8 ... ok
test lib::rsa::round_trip_sign_verification_pem_pkcs1 ... ok
test lib::eddsa::ed_jwk ... ok
test lib::eddsa::ed_x ... ok
test lib::eddsa::round_trip_claim ... ok
test lib::eddsa::round_trip_sign_verification_pem ... ok
test lib::eddsa::round_trip_sign_verification_pk8 ... ok
test lib::header::x5c_der_invalid_chain ... ok
test lib::header::x5c_der_valid_chain ... ok
test lib::header::x5c_der_empty_chain ... ok

test result: ok. 21 passed; 0 failed; 0 ignored
```

### DIFF
--- a/tests/rsa/mod.rs
+++ b/tests/rsa/mod.rs
@@ -3,6 +3,7 @@ use jsonwebtoken::{
     Algorithm, DecodingKey, EncodingKey,
 };
 use serde::{Deserialize, Serialize};
+use wasm_bindgen_test::wasm_bindgen_test;
 
 #[cfg(feature = "use_pem")]
 use jsonwebtoken::{decode, encode, Header, Validation};
@@ -27,7 +28,9 @@ pub struct Claims {
 
 #[cfg(feature = "use_pem")]
 #[test]
+#[wasm_bindgen_test]
 fn round_trip_sign_verification_pem_pkcs1() {
+
     let privkey_pem = include_bytes!("private_rsa_key_pkcs1.pem");
     let pubkey_pem = include_bytes!("public_rsa_key_pkcs1.pem");
     let certificate_pem = include_bytes!("certificate_rsa_key_pkcs1.crt");
@@ -58,6 +61,7 @@ fn round_trip_sign_verification_pem_pkcs1() {
 
 #[cfg(feature = "use_pem")]
 #[test]
+#[wasm_bindgen_test]
 fn round_trip_sign_verification_pem_pkcs8() {
     let privkey_pem = include_bytes!("private_rsa_key_pkcs8.pem");
     let pubkey_pem = include_bytes!("public_rsa_key_pkcs8.pem");
@@ -88,6 +92,7 @@ fn round_trip_sign_verification_pem_pkcs8() {
 }
 
 #[test]
+#[wasm_bindgen_test]
 fn round_trip_sign_verification_der() {
     let privkey_der = include_bytes!("private_rsa_key.der");
     let pubkey_der = include_bytes!("public_rsa_key.der");
@@ -103,6 +108,7 @@ fn round_trip_sign_verification_der() {
 
 #[cfg(feature = "use_pem")]
 #[test]
+#[wasm_bindgen_test]
 fn round_trip_claim() {
     let my_claims = Claims {
         sub: "b@b.com".to_string(),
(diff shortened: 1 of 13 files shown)


tests/rsa/mod.rs
```diff
@@ -201,6 +209,7 @@ fn rsa_jwk() {
 // https://jwt.io/ is often used for examples so ensure their example works with jsonwebtoken
 #[cfg(feature = "use_pem")]
 #[test]
+#[wasm_bindgen_test]
```
Keats, line 212: Do we really need to put that on every test? :(

Keats (approved):

tests/rsa/mod.rs
lpotthast: I think this is unavoidable. Couldn't find aything about the reusability of the standard `#[test]` macro.

Keats:
>  I do not know if unit-test should be kept out though, and running integration-tests only is enough for the wasm case..

I think only on integration tests. No need to run unit tests i guess. Although we should probably only have that attribute on wasm target to avoid adding compile time/deps when not needed?

Keats:
Fixed the CI, thanks!

PR #351, opened 2023-12-11
Make family public

brymko:
A patch to hopefully fix all the AlgorithmFamily confusion. We ran into a very similar issue which i think are many describing here. But i noticed that _we_ specifically like to pin the validation of the jwt to the decoding/encoding key that is being used. But unfortunately this is currently not possible without adding more state tracking for our keys, which is annoying given that they already store this information. 

This patch will publicize a method to get the `AlgorithmFamily` from both encoding & decoding key. And also adds a method to AlgorithmFamily  that will return a list of all algorithms which are part of the family

### DIFF
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -11,6 +11,7 @@ pub enum AlgorithmFamily {
 }
 
 impl AlgorithmFamily {
+    /// A list of all possible Algorithms that are part of the family.
     pub fn algorithms(&self) -> &[Algorithm] {
         match self {
             Self::Hmac => &[Algorithm::HS256, Algorithm::HS384, Algorithm::HS512],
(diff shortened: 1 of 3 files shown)


Keats:
Can you add a snippet of how you would use that?

brymko:
We're using the following primitives to setup correct validation and encoding parameters based on the key given:

```rust
fn validation_from_decoding_key(key: &jsonwebtoken::DecodingKey) -> jsonwebtoken::Validation {
    let mut ret = jsonwebtoken::Validation::default();

    ret.algorithms = key.family().algorithms().to_vec();

    ret
}

fn header_from_encoding_key(key: &jsonwebtoken::EncodingKey) -> jsonwebtoken::Header {
    jsonwebtoken::Header {
        alg: *key
            .family()
            .algorithms()
            .first()
            .expect("No algorithms found there must be atleast one algorithm per family"),
        ..jsonwebtoken::Header::default()
    }
}

```

Issue #357, opened 2024-01-09
EncodingKey to JWK


andrewbaxter:
The same way that there's a way to go from JWK to DecodingKey, when creating a JWT you generate a key and need to embed information about that key in the payload.

I think the expected usage (? I couldn't find an example here) is to generate the private key and at the same point generate the JWK, then pass them around together.  This risks mixing up the JWK and key leading to invalid tokens, vs generating the JWK from the key at the time of generating the JWT.

Right now EncodingKey contains an opaque u8 blob which is decoded into the correct key type at the time of signing.  Following this approach, it would have to be decoded twice, once for signing and once for JWK generation.  I think it may be better to decode the key into the parsed structure immediately when creating the EncodingKey -- this has the added benefit of validating the key early.

andrewbaxter:
Ah... ring limitations. https://github.com/briansmith/ring/issues/579 and confirmed in https://github.com/breard-r/acmed/issues/2#issuecomment-505802712

andrewbaxter:
Partially solved: https://github.com/briansmith/ring/issues/1161 (RSA)

andrewbaxter:
It looks like it's pretty easy to get the ecdsa public key coords, if uncompressed it's just 1 byte past the public key start. https://datatracker.ietf.org/doc/html/rfc5480#section-2.2 with a hint in https://datatracker.ietf.org/doc/html/rfc7518#section-6.2.1.2 and .3

Ed25519 looks very complex, comparatively: https://crypto.stackexchange.com/questions/51663/how-can-i-split-a-packed-ed25519-public-signing-key-into-its-x-and-y-coordinates

andrewbaxter:
FWIW Caddy doesn't support Ed25519 either: https://github.com/mholt/acmez/blob/422d0ad3b73253c35dc2b9d0be29ad55385df3ff/examples/attestation/main.go#L268

PR #360, opened 2024-01-10
Method to get JWK from EncodingKey

andrewbaxter:
Per #357 

I excluded Ed25519 for now, but the rest works as expected I believe.  I'm going to test it with the code I wanted to use it in originally for now.


andrewbaxter:
My (limited) real world testing worked.

Obviously splitting the bytes is a hack, but I wasn't able to find the ring internal methods that extract that data.  I may continue to look, but I think this is _relatively_ safe.

andrewbaxter:
https://github.com/andrewbaxter/fork-jsonwebtoken/pull/2 targetted at the ACME/JWS branch

PR #363, opened 2024-01-14
Add support for JWK thumbprints

andrewbaxter:
https://datatracker.ietf.org/doc/html/rfc7638

I republished the ring digests so that the feature could be used without matching ring versions.

The thumbprint json is hardcoded templates because I'm not sure if serde_json guarantees json string format (ordering, spacing, etc). The spec is written in such a way that quotation marks and backslashes won't appear in the serialized values so there's no risk of escape issues.

I think the readme's getting a bit crowded, but  I didn't want to introduce unrelated changes here so I didn't touch it.

Aside from the unit test, I was testing with letsencrypt/pebble with EC keys and it seemed to be okay (although I switched from serde_json to hardcoded string templates after that).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,3 +22,7 @@ pub use decoding::{decode, decode_header, DecodingKey, TokenData};
 pub use encoding::{encode, EncodingKey};
 pub use header::Header;
 pub use validation::{get_current_timestamp, Validation};
+
+pub use ring::digest::SHA256 as DIGEST_SHA256;
+pub use ring::digest::SHA384 as DIGEST_SHA384;
+pub use ring::digest::SHA512 as DIGEST_SHA512;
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -22,3 +22,7 @@ pub use decoding::{decode, decode_header, DecodingKey, TokenData};
 pub use encoding::{encode, EncodingKey};
 pub use header::Header;
 pub use validation::{get_current_timestamp, Validation};
+
+pub use ring::digest::SHA256 as DIGEST_SHA256;
```
Keats, line 26: Let's not expose ring at all. Make an enum with hidden internals and let users use that.

andrewbaxter: Done!

andrewbaxter:
Targetted at the ACME/JWS branch here https://github.com/andrewbaxter/fork-jsonwebtoken/pull/1

PR #367, opened 2024-01-20
Refactor auth0 example

tottoto:
Decouples auth0 examples' pattern matching to the more flattened structure, and uses header's `alg` for specifying the algorithm.

### DIFF
--- a/examples/auth0.rs
+++ b/examples/auth0.rs
@@ -1,9 +1,8 @@
 /// Example for the backend to backend implementation
 use std::collections::HashMap;
-use std::str::FromStr;
 
-use jsonwebtoken::jwk::AlgorithmParameters;
-use jsonwebtoken::{decode, decode_header, jwk, Algorithm, DecodingKey, Validation};
+use jsonwebtoken::jwk::{AlgorithmParameters, JwkSet};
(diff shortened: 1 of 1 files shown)


examples/auth0.rs
```diff
@@ -1,44 +1,42 @@
...
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let jwks: jwk::JwkSet = serde_json::from_str(JWKS_REPLY).unwrap();
+    let jwks: JwkSet = serde_json::from_str(JWKS_REPLY).unwrap();
+    let header = decode_header(TOKEN).unwrap();
 
-    let header = decode_header(TOKEN)?;
-    let kid = match header.kid {
-        Some(k) => k,
-        None => return Err("Token doesn't have a `kid` header field".into()),
+    let Some(kid) = header.kid else {
```
Keats, line 16: Since which version of Rust is that stable? First time I'm seeing that syntax

tottoto: let-else statements is stabilized at Rust 1.65.

https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#let-else-statements

PR #377, opened 2024-03-12
Add feature based on aws-lc-rs cryptographic library instead of ring

KirilNN:
The goal here is to enable feature based on [aws-lc-rs](https://crates.io/crates/aws-lc-rs/1.6.2) library, so it can be used on demand instead of ring which is not FIPS certified.


Cargo.toml
```diff
@@ -50,6 +51,7 @@ criterion = { version = "0.4", default-features = false }
 [features]
 default = ["use_pem"]
 use_pem = ["pem", "simple_asn1"]
+aws_lc_rs = ["aws-lc-rs"]
```
Keats, line 54: can we just call it `fips`? That lib name is not obvious

Keats (changes requested):
Can you add running with this feature on to the CI?

KirilNN:
@Keats thanks for taking a look, I changed the feature name and added it to the CI, do you mind taking a look again?

Keats:
Have you seen this PR: https://github.com/Keats/jsonwebtoken/pull/318 ? It's something I was considering

KirilNN:
Well this is also reasonable imo, for me the question is whether these libs are FIPS compliant, cause this is the biggest issue we are tackling with this PR, as AWS did get it certified. Can we do features as well - like remove ring, get a feature for FIPS using aws lib and all the rest with the lib combo? Does it sound good? I believe removing ring is OK for everyone.

Keats:
It looks like building this lib on windows is problematic? https://github.com/aws/aws-lc/issues/1477

KirilNN:
It looks like the symcrypt from Microsoft does not have wide support of platforms, although has the needed features, so I guess we can either merge the PR you suggest which is not guaranteed we have FIPS compliance, or we wait for the AWS folks to fix the windows build?

GlenDC:
@Keats / @KirilNN I have some bandwidth to spare and in light of #399 I can help maintenance given this is the JWT lib with the most future benefit.

GlenDC:
@Keats I assume it would make sense to first focus on https://github.com/Keats/jsonwebtoken/pull/410 and than take on this? or how do you see this?

Keats:
https://github.com/Keats/jsonwebtoken/pull/410 will make this PR obsolete

GlenDC:
> #410 will make this PR obsolete

Cool, in that case I'll focus on rebooting #410, feel free to already close this, unless you have a good reason not to.

PR #380, opened 2024-03-26
Update MSRV in README.md to match Cargo.toml

nerdile:
Fix the readme to match the required Rust version (MSRV) in Cargo.toml.


Keats:
Should we just link to the line in the Cargo.toml? Keeping them in sync seem like it would never work

PR #397, opened 2024-07-04
 Update base64 to 0.22 

maflcko:
According to https://github.com/marshallpierce/rust-base64/blob/master/RELEASE-NOTES.md this does not affect this project. (Only change is a speedup, a minor change of the `InvalidLength` error, and a fixed `BIN_HEX`)

Also in a separate doc-only commit, address https://github.com/Keats/jsonwebtoken/pull/380#issuecomment-2030469801 to fix #380.


Keats:
It's released

rimutaka:
Relates to #411

PR #410, opened 2024-10-03
Decoupled crypto backends

sidrubs:
# Introduction

This is a POC at abstracting the signing and verifying algorithms behind the [`JwtSigner`](https://github.com/sidrubs/jsonwebtoken/blob/decoupled-crypto-backends/src/crypto/mod.rs#L17) and [`JwtVerifier`](https://github.com/sidrubs/jsonwebtoken/blob/decoupled-crypto-backends/src/crypto/mod.rs#L25) traits whilst not breaking the `encode` and `decode` public interfaces that people know and love.

This allows any crypto backend to be used, as long as it implements the `JwtSigner` and `JwtVerifier` traits. Allowing developers to provide their own backend dependent on their use case. This should also enable more flexibility for the `jsonwebtoken` crate to prepackage different backends (e.g. `ring`, `aws_lc_rs`, and `RustCrypto`).

# Implementation

[`JwtEncoder`](https://github.com/sidrubs/jsonwebtoken/blob/decoupled-crypto-backends/src/encoding.rs#L160-279)and [`JwtDecoder`](https://github.com/sidrubs/jsonwebtoken/blob/decoupled-crypto-backends/src/decoding.rs#L284-419) builder style clients were added to the public API allow the developer to use the included algorithms or provide their own.

The original [`encode`](https://github.com/sidrubs/jsonwebtoken/blob/decoupled-crypto-backends/src/encoding.rs#L123) and [`decode`](https://github.com/sidrubs/jsonwebtoken/blob/decoupled-crypto-backends/src/decoding.rs#L217) public interfaces call this builder under-the-hood with a factory function choosing the correct algorithm backend.

> **Note: ** This POC only [implements](https://github.com/sidrubs/jsonwebtoken/blob/decoupled-crypto-backends/src/crypto/hmac.rs) the `HMAC` family of algorithms as I didn't want to spend a bunch of time on something that was not going to be accepted.

# Advantages

- The builder-style API provides compile time checks to the algorithm and secret's compatibility
- A developer is able to provide their own backend so does not need to rely on the `jsonwebtoken` crate maintainers to add something for them.
- Allows the crate to be flexible, thus hopefully reduce fragmentation in the ecosystem.

# Disadvantages

- I have tried to make this backwards compatible, but this introduces some little idiosyncrasies, especially on the validation side (making use of the original `Validation` struct). Everything works, it would just seem a little weird to someone looking at parts of the codebase without context. 

# Wrapping Up

I understand that this a a pretty big change architecturally to the `jsonwebtoken` crate, so fully understand if you don't want to go this route. I do feel, however, that it makes the project more flexible and maintainable going forward.

Any suggestions are welcome.


src/crypto/hmac.rs
```diff
@@ -1,106 +1,161 @@
...
+    }
+}
 
-    fn verify(&mut self, signature: &[u8], message: &[u8]) -> bool {
-        self.reset();
-        self.update(message);
-        self.clone().verify_slice(signature).is_ok()
+impl JwtVerifier for Hs512 {
+    fn algorithm(&self) -> Algorithm {
+        Algorithm::HS512
```
Keats, line 159: Looks like the kind of code that could be simplified with macros

src/decoding.rs
```diff
@@ -259,14 +219,48 @@ pub fn decode<T: DeserializeOwned>(
...
 ) -> Result<TokenData<T>> {
-    match verify_signature(token, key, validation) {
-        Err(e) => Err(e),
-        Ok((header, claims)) => {
-            let decoded_claims = DecodedJwtPartClaims::from_jwt_part_claims(claims)?;
-            let claims = decoded_claims.deserialize()?;
-            validate(decoded_claims.deserialize()?, validation)?;
-
-            Ok(TokenData { header, claims })
+    // The decoding step is unfortunately a little clunky but that seems to be how I need to do it to fit it into the `decode` function
```
Keats, line 222: how would it look like without that constraint?

Keats (commented):
I like the look of it but would like both RustCrypto and aws-lc support to make sure it works well.
I'm not too sure about the builders but I guess we can discuss that later

sidrubs:
> I like the look of it but would like both RustCrypto and aws-lc support to make sure it works well. I'm not too sure about the builders but I guess we can discuss that later

Oh yeah sure thing. I'll add `aws_lc_rs` `JwtSigner` and `JwtVerifier` trait implementations (selected by a feature flag).

With regards to the builder, here was my thought process:

The `JwtSigner` and `JwtVerifier` traits have an `algorithm` method so that we know which algorithm the `JwtSigner` supports. I felt that most people will only ever use the a default header (with the correct algorithm), so why not automatically set it. But to be backwards compatible with the current `encode` function we would need to be able to set a fully custom header if needs be.

We could avoid the builder pattern by creating a function like the following and calling it from the current `encode` function (and initializing the `signing_provider` based on the `header` and the `key`. This however would not check the algorithm and secret at compile time, but that is the current behavior, so no problem.

```rs
pub fn generic_encode<S: JwtSigner, T: Serialize>(
    signing_provider: S,
    header: &Header,
    claims: T,
    key: &EncodingKey
) -> Result<String> {
    ...
}
```

src/crypto/hmac.rs
sidrubs: For repeated things that that, sure it could :+1: 

I feel that we are only repeating it a couple times so I don't believe the complexity of a macro is worth it. I also find it easier to read normal Rust code. But maybe that is just me :shrug:

src/decoding.rs
sidrubs: Hmm, I think that comment doesn't really apply anymore (I should have removed it). There was a stage that I was less sure what was going on so I had a bit of a mess in there :facepalm:

Keats:
I think we can have both the builder and the function?

sidrubs:
> I think we can have both the builder and the function?

I just want to understand correctly as there are two functions I have talked about and  I feel that we might be talking past each other.

1. There are the original [`encode`](https://github.com/Keats/jsonwebtoken/blob/master/src/encoding.rs#L121) and [`decode`](https://github.com/Keats/jsonwebtoken/blob/master/src/decoding.rs#L257) functions that are currently part of the public API in the `jsonwebtoken` crate. In my POC, I kept these function signatures as they were (so as not to break current the current public API), but they actually called the builder under-the-hood.
2. The second function came about when you mentioned that maybe the builder might not be ideal. So instead of using a builder, I [suggested](https://github.com/Keats/jsonwebtoken/pull/410#issuecomment-2392466079) that we could use a function that allowed us to make use of arbitrary crypto backends. This would be called under-the-hood of the original `encode` and `decode` functions.

I both cases I was visualizing that we keep the same functionality of the original functions [1], but the internal business logic of these functions would either be provided by a builder or a function [2] style interface. So in addition to the original functions [1], we would also expose a method that an arbitrary backend could be used (this would either be a builder or another function).

Does this clear things up?

sidrubs:
I have added an `aws-lc-rs` implementation behind a feature flag (again, only HMAC for now until we agree on a way going forward).

Currently there are `rust_crypto` and `aws_lc_rs` feature flags (more can be added easily). They will cause a descriptive compile error if enabled at the same time, but I can change this behavior to a run-time selection if needed.

Keats:
> I both cases I was visualizing that we keep the same functionality of the original functions [1], but the internal business logic of these functions would either be provided by a builder or a function [2] style interface. So in addition to the original functions [1], we would also expose a method that an arbitrary backend could be used (this would either be a builder or another function).
> Does this clear things up?

I think that's fine, as long as we can keep the keys in once_cell/lazy_static and use them without having to clone anything

src/decoding.rs
```diff
@@ -286,3 +286,140 @@ pub fn decode_header(token: &str) -> Result<Header> {
...
+/// let hmac_secret = HmacSecret::from_secret(b"secret");
+///
+/// let claims = JwtDecoder::hs_256(hmac_secret)
+///     .unwrap()
+///     .decode::<Claims>(&token)
+///     .unwrap();
+/// ```
+pub struct JwtDecoder {
+    verifying_provider: Box<dyn JwtVerifier>,
+    validation: Validation,
```
Keats, line 316: Should it be `Option<Validator>`? The default Validation is probably not useful for a lot of cases so might as well not have one

src/encoding.rs
```diff
@@ -122,10 +130,156 @@ pub fn encode<T: Serialize>(header: &Header, claims: &T, key: &EncodingKey) -> R
...
+    /// Create a new [`JwtEncoder`] with any `signing_provider` that implements the [`JwtSigner`] trait.
+    pub fn from_signer<S: JwtSigner + 'static>(signing_provider: S) -> Self {
+        Self::from_boxed_signer(Box::new(signing_provider))
+    }
+
+    /// Create a new [`JwtEncoder`] with any `signing_provider` that implements the [`JwtSigner`] trait.
+    pub fn from_boxed_signer(signing_provider: Box<dyn JwtSigner>) -> Self {
+        // Determine a default header
+        let mut header = Header::new(signing_provider.algorithm());
+        header.typ = Some("JWT".to_owned());
```
Keats, line 207: That's the default I believe?

Keats (commented):
I like where it's going

src/encoding.rs
sidrubs: Yes, you are correct. I'll fix that up.

src/decoding.rs
sidrubs: I am currently reusing a lot of the current validation logic that makes use of fields in the `Validation` struct. So in the initializer for the `JwtDecoder` it is [creating](https://github.com/sidrubs/jsonwebtoken/blob/decoupled-crypto-backends/src/decoding.rs#L327) a "default" `Validation` struct based on the algorithm of the `verifying_provider`. This is then used for the validation logic. If we were to use an `Option<Validator>` we would still need to provide default values to use somewhere in the code.

If someone wants to provide custom validation logic they can call the `with_validation` builder style method to replace the "default" validation that is generated in the initializer.

sidrubs: If one wanted to clean things up one could flatten the `Validation` fields and methods into the `JwtDecoder`, but because we want to keep the original `decode` function working the `Validation` code needs to be there, so might as well make use of it.

sidrubs:
It seems that this POC might be used. So I'll go ahead and add the rest of the algorithms, keeping in mind some of the comments above, and get everything up to scratch to pass CI.

Unless there are any other concerns?

src/decoding.rs
Keats: We should have the user pass it explicitly to the constructor then, implicit default validation is not good.

sidrubs: I think I should replace the builder with a function (see [here](https://github.com/Keats/jsonwebtoken/pull/410#issuecomment-2392466079)). This would remove these smaller complexities.

Keats:
Go ahead, we can play with the DX once it's implemented. Maybe use macros to implement them so it's easy to make changes if needed rather than copy/paste

sidrubs:
I have converted the builder to [`_encode` ](https://github.com/sidrubs/jsonwebtoken/blob/decoupled-crypto-backends/src/encoding.rs#L144)and[ `_decode`](https://github.com/sidrubs/jsonwebtoken/blob/decoupled-crypto-backends/src/decoding.rs#L250) functions that are called by the original public facing `encode` and `decode` functions. I think it does simplify things a bit.

I am about half done adding the RSA algorithms functionality.

I did however recently notice that the original [`sign`](https://github.com/Keats/jsonwebtoken/blob/master/src/crypto/mod.rs#L25)and [`verify`](https://github.com/Keats/jsonwebtoken/blob/c733c786fc9a31362ad21a7498cbd926d1aa7e4b/src/crypto/mod.rs#L68) functions in the `crypto` module are `pub`. The "new" trait based architecture does not make use of them, should I recreate their functionality using the traits based approach so that we don't break the current public facing API? Or would you envision this being a major version bump and removing those functions?

Keats:
They are used by some people who only care about the crypto primitives so I would re-implement them and keep them pub

Keats:
Hi @sidrubs will you have time to finish that PR?

sidrubs:
> Hi @sidrubs will you have time to finish that PR?

Hey @Keats we have slightly changed our focus, so am not going to finish the PR anytime soon :-/

GlenDC:
@Keats

> **Note: ** This POC only [implements](https://github.com/sidrubs/jsonwebtoken/blob/decoupled-crypto-backends/src/crypto/hmac.rs) the HMAC family of algorithms as I didn't want to spend a bunch of time on something that was not going to be accepted.

Would this scope still be acceptable or do you need them all to be implemented in 1 PR?

Keats:
I think it would be easier to implement all of it in one PR

GlenDC:
> I think it would be easier to implement all of it in one PR

Perfect. Thanks for verifying!

Keats:
@GlenDC any updates?

kamulos:
The situation with the `ring` crate is a bit worrying from the outside: in the last days it went to unmaintained status, changed hands, changed hands back.

So big :+1: for the direction this pr is going into. Personally I will be using the `aws-lc` backend because `rustls` uses it anyways.

pauliyobo:
Hello
@Keats what is the status of the PR? That is to say what would be needed in order to merge it?
It would seem like hmac and rsa support was added for aws-lc-rs but only hmac support was added for rust-crypto.
And in order to help, would it be better to open a new PR basing the work on this?
Thanks.

Keats:
The status is what you see in this PR. This is definitely the way forward and if people want to do PRs on that PR to finish it, please go ahead!

PR #416, opened 2025-02-10
Derive Default for JwkSet

tristan957:
Useful for creating empty an empty JWKS.

### DIFF
--- a/src/jwk.rs
+++ b/src/jwk.rs
@@ -419,7 +419,7 @@ impl Jwk {
 }
 
 /// A JWK set
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


tristan957:
> What would be the use case for this change?

I want to send an empty JwkSet for whatever reason.

```diff
- let empty_set = JwkSet {
-   keys: Vec::default(),
- };
+ let empty_set = JwkSet::default();
```

tristan957:
> @tristan957 , I feel like this change is setting someone for failure down the track. An empty JwkSet is a source of potential bugs - the compiler will not be able to reason because the struct is there, but the keys are not.
> 
> Making this explicit during the construction of the set feels safer to me.

I'm not sure how someone could be set up for failure. I am requesting an empty JwkSet explicitly. What are the potential bugs here?

Issue #401, opened 2024-08-26
Provide a better API for decoding a token without signature validation


tyilo:
Currently you have to use:
```rust
// Algorithm can be arbitrarily chosen
let mut validation = jsonwebtoken::Validation::new(jsonwebtoken::Algorithm::RS256);
validation.insecure_disable_signature_validation();

// Key can be arbitrarily chosen
let key = DecodingKey::from_secret(&[]);
let payload = jsonwebtoken::decode::<Claims>(token, &key, &validation).unwrap();
```

I think the following API would be better:
```rust
let mut validation = jsonwebtoken::Validation::insecure_without_signature_validation();
let payload = jsonwebtoken::insecure_decode_without_signature_validation::<Claims>(token, &validation).unwrap();
```

You avoid having to choose a random algorithm and decoding key that isn't ever used.

Keats:
Honestly, decoding a token without validating the signature is something that you shouldn't do most of the time so I do not particularly care about making it user friendly

tyilo:
> Honestly, decoding a token without validating the signature is something that you shouldn't do most of the time so I do not particularly care about making it user friendly

It is useful as a client using the token to be able to see what claims are inside the token. `exp` can be really useful.

Keats:
Well you can't really trust any of the things you see in the claims unless you validate the signature

pooriatgh:
Today, I encountered the same issue. It would be nice if we could make this easier. Here is the scenario:

1. I need to decode the token, look into the claims without validation, and extract the issuer's name.
2. Use the issuer's name to find the correct public key file from a list of public keys I have.
3. Use the key to perform the verification/validation.

I tried the solution provided by @tyilo , but I failed to decode without the audience validation step. I used this option to disable it:


```rust
let mut validation = jsonwebtoken::Validation::new(jsonwebtoken::Algorithm::RS256);
validation.insecure_disable_signature_validation();
validation.validate_aud = false;
```

For example in typescript this package provide more or less what I need: [node-jsonwebtoken](https://github.com/auth0/node-jsonwebtoken)

ppamorim:
> > Well you can't really trust any of the things you see in the claims unless you validate the signature
> 
> Sure, but I'm the client. I don't have access to the server's secret key.

[...]

PR #418, opened 2025-02-16
feat: Add insecure_decode_without_signature_validation

Nereuxofficial:
As discussed [here](https://github.com/Keats/jsonwebtoken/issues/401#issuecomment-2661123916) adds support for insecure_decode_without_serialization as well as a doc test for it. I also fixed some doc warnings while at it.


src/decoding.rs
```diff
@@ -286,3 +286,43 @@ pub fn decode_header(token: &str) -> Result<Header> {
...
+/// }
+///
+/// // Example token from jwt.io
+/// let token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOjEyMzQ1Njc4OTAsIm5hbWUiOiJKb2huIERvZSIsImlhdCI6MTUxNjIzOTAyMiwiZXhwIjoyNTE2MjM5MDYwfQ.Yf3kCk-BdkW3DZNao3lwMoU41ujnt86OgewBA-Q2uBw".to_string();
+/// let validation = Validation::new(Algorithm::HS256);
+/// let claims = insecure_decode_without_signature_validation::<Claims>(&token, &validation).unwrap();
+/// ```
+pub fn insecure_decode_without_signature_validation<T: DeserializeOwned>(
+    token: &str,
+    validation: &Validation,
```
rimutaka, line 320: I'm missing something here. Please ignore if I'm wrong.
The function says it does not do validation and the algorithm can be arbitrary chose, yet, there is a validation param and the validation step. So it's kind of partial validation.

I find it confusing and is not what #401 calls for.

src/decoding.rs
```diff
@@ -286,3 +286,43 @@ pub fn decode_header(token: &str) -> Result<Header> {
...
+) -> Result<T> {
+    let (_, rest) = expect_two!(token.rsplitn(2, '.'));
+    let (claims, _) = expect_two!(rest.rsplitn(2, '.'));
+    let decoded_claims = DecodedJwtPartClaims::from_jwt_part_claims(claims)?;
+    let claims = decoded_claims.deserialize()?;
+    validate(decoded_claims.deserialize()?, validation)?;
```
rimutaka, line 326: The function description says "the claims are not validated", but they are.

src/decoding.rs
```diff
@@ -286,3 +286,43 @@ pub fn decode_header(token: &str) -> Result<Header> {
...
+/// // Example token from jwt.io
+/// let token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOjEyMzQ1Njc4OTAsIm5hbWUiOiJKb2huIERvZSIsImlhdCI6MTUxNjIzOTAyMiwiZXhwIjoyNTE2MjM5MDYwfQ.Yf3kCk-BdkW3DZNao3lwMoU41ujnt86OgewBA-Q2uBw".to_string();
+/// let validation = Validation::new(Algorithm::HS256);
+/// let claims = insecure_decode_without_signature_validation::<Claims>(&token, &validation).unwrap();
+/// ```
+pub fn insecure_decode_without_signature_validation<T: DeserializeOwned>(
+    token: &str,
+    validation: &Validation,
+) -> Result<T> {
+    let (_, rest) = expect_two!(token.rsplitn(2, '.'));
```
rimutaka, line 322: JWT has 3 parts: Header, Payload, Signature.

The existing code refers to them by those names:
```
    let (signature, message) = expect_two!(token.rsplitn(2, '.'));
    let (payload, header) = expect_two!(message.rsplitn(2, '.'));
```

It would be good to keep them consistent throughout the file.

rimutaka (commented):
Is it a good idea to add tests?
- valid signature
- invalid signature
- invalid token

I'm not a maintainer, so feel free to ignore.

src/decoding.rs
Nereuxofficial: Ah i was not completely sure that was the case as the creator of the issue wanted essentially this API:
``rs
let mut validation = jsonwebtoken::Validation::insecure_without_signature_validation();
let payload = jsonwebtoken::insecure_decode_without_signature_validation::<Claims>(token, &validation).unwrap();
```
And IMO then a Validation::insecure_none or similar should exist to make this function more general.
Nonetheless the Doc comment is wrong so thanks for pointing that out!

bombsimon:
> > Can I help out to move this forward?
> 
> Feel free to do so as i no longer have a use-case for this feature and the envisioned API is a bit unclear.

It looks like the easiest way would just be the new API that just accepts the token and does identical decoding part as today but without verifying any signatures.

I don't want to PR anything since you have this going but [here is my take](https://github.com/Keats/jsonwebtoken/compare/master...bombsimon:jsonwebtoken:add-insecure-decode?expand=1) which is simply just the new `insecure_decode_without_signature_validation` that accepts a token and some tests for it.

Feel free to grab any parts of that if you want to but would be nice if we could land this.

Issue #368, opened 2024-01-23
How to customize the Header field


jiashiwen:
Some of my python or java programs found the following definitions

python

```
  return jwt.encode(
    payload,
    secret,
    algorithm="HS256",
    headers={"alg": "HS256", "sign_type": "SIGN"},
)
```

java

```
 headerClaims.put("alg", "HS256");
 headerClaims.put("sign_type", "SIGN");
 String token = JWT.create().withPayload(payload).withHeader(headerClaims).sign(alg);
```

However, the "sign_type" field is not seen in the jsonwebtoken::Header definition. How to customize the header's custom fields?

Keats:
You can't, you can only use header fields from the spec in this library

jiashiwen:
A hashmap may be needed to provide the function of custom fields, and some access may require

hezhaozhao-git:
+1

yspreen:
very unfortunate, makes the library unusable for cases like https://developer.apple.com/documentation/weatherkitrestapi/request_authentication_for_weatherkit_rest_api

which needs an `id` field in the header

arckoor:
`Header.extras` allows this

PR #420, opened 2025-02-24
feat: allow users to specify a custom header not defined in the struct

jayakasadev:
* allow users to specify custom header values for token generation
* remove deprecated verify_slices_are_equal
* update tests

### DIFF
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -1,3 +1,4 @@
+use ring::constant_time::verify_slices_are_equal;
 use ring::{hmac, signature};
 
 use crate::algorithms::Algorithm;
@@ -9,16 +10,6 @@ use crate::serialization::{b64_decode, b64_encode};
 pub(crate) mod ecdsa;
 pub(crate) mod eddsa;
 pub(crate) mod rsa;
-
-/// Only used internally when signing/verifying with HMAC, to map from our enum to the Ring HMAC structs.
-fn alg_to_hmac(alg: Algorithm) -> hmac::Algorithm {
-    match alg {
-        Algorithm::HS256 => hmac::HMAC_SHA256,
-        Algorithm::HS384 => hmac::HMAC_SHA384,
-        Algorithm::HS512 => hmac::HMAC_SHA512,
-        _ => unreachable!("Tried to get HMAC alg for a non-HMAC algorithm"),
-    }
-}
 
 /// The actual HS signing + encoding
 /// Could be in its own file to match RSA/EC but it's 2 lines...
@@ -33,9 +24,9 @@ pub(crate) fn sign_hmac(alg: hmac::Algorithm, key: &[u8], message: &[u8]) -> Str
 /// If you just want to encode a JWT, use `encode` instead.
 pub fn sign(message: &[u8], key: &EncodingKey, algorithm: Algorithm) -> Result<String> {
     match algorithm {
-        Algorithm::HS256 | Algorithm::HS384 | Algorithm::HS512 => {
-            Ok(sign_hmac(alg_to_hmac(algorithm), key.inner(), message))
-        }
+        Algorithm::HS256 => Ok(sign_hmac(hmac::HMAC_SHA256, key.inner(), message)),
+        Algorithm::HS384 => Ok(sign_hmac(hmac::HMAC_SHA384, key.inner(), message)),
+        Algorithm::HS512 => Ok(sign_hmac(hmac::HMAC_SHA512, key.inner(), message)),
 
         Algorithm::ES256 | Algorithm::ES384 => {
             ecdsa::sign(ecdsa::alg_to_ec_signing(algorithm), key.inner(), message)
@@ -83,10 +74,8 @@ pub fn verify(
     match algorithm {
         Algorithm::HS256 | Algorithm::HS384 | Algorithm::HS512 => {
             // we just re-sign the message with the key and compare if they are equal
-            let encoding_key = &EncodingKey::from_secret(key.as_bytes());
-            let key = &hmac::Key::new(alg_to_hmac(algorithm), encoding_key.inner());
-            let digest = hmac::sign(key, message);
-            Ok(hmac::verify(key, message, digest.as_ref()).is_ok())
+            let signed = sign(message, &EncodingKey::from_secret(key.as_bytes()), algorithm)?;
+            Ok(verify_slices_are_equal(signature.as_ref(), signed.as_ref()).is_ok())
         }
         Algorithm::ES256 | Algorithm::ES384 => verify_ring(
             ecdsa::alg_to_ec_verification(algorithm),
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -1,15 +1,35 @@
-use ring::constant_time::verify_slices_are_equal;
 use ring::{hmac, signature};
 
 use crate::algorithms::Algorithm;
 use crate::decoding::{DecodingKey, DecodingKeyKind};
 use crate::encoding::EncodingKey;
-use crate::errors::Result;
+use crate::errors::{Error, Result};
 use crate::serialization::{b64_decode, b64_encode};
 
 pub(crate) mod ecdsa;
 pub(crate) mod eddsa;
 pub(crate) mod rsa;
+
+/// Only used internally when signing/verifying with HMAC, to map from our enum to the Ring HMAC structs.
+fn alg_to_hmac(alg: Algorithm) -> hmac::Algorithm {
+    match alg {
+        Algorithm::HS256 => hmac::HMAC_SHA256,
+        Algorithm::HS384 => hmac::HMAC_SHA384,
+        Algorithm::HS512 => hmac::HMAC_SHA512,
+        _ => unreachable!("Tried to get HMAC alg for a non-HMAC algorithm"),
+    }
+}
+
+/// Returns `Ok(())` if `a == b` and `Err(error::Unspecified)` otherwise.
+pub fn verify_slices_are_equal(a: &[u8], b: &[u8]) -> Result<()> {
+    if b.len() != a.len() {
+        return Err(Error::from(ring::error::Unspecified));
+    }
+    match openssl::memcmp::eq(a, b) {
+        true => Ok(()),
+        _ => Err(Error::from(ring::error::Unspecified)),
+    }
+}
 
 /// The actual HS signing + encoding
 /// Could be in its own file to match RSA/EC but it's 2 lines...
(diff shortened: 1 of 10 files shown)


rimutaka:
@jayakasadev , what is the use case for this?
Is `extra` a common name for this?
I could not find any references or examples for this in other implementations.

jayakasadev:
I have a client which requires that I pass in some custom headers in the token. The `extra` field allows users to optionally add such headers.

rimutaka:
@jayakasadev , are those headers essential to decoding and validating the token?
Can you provide a real life example?

jayakasadev:
example: https://docs.cdp.coinbase.com/advanced-trade/docs/ws-auth
coinbase's auth requires a nonce header in the jwt token

rimutaka:
@Keats , looks like there is an uncommon, but legit use case for this.
Are you open to merging it?
If yes, I'm happy to work with @jayakasadev to get it over the line (docs, probably more tests, code tidy up).

jayakasadev:
@rimutaka let me know what changes are needed

Keats:
Looks like clippy/cargo fmt needs to be ran

src/header.rs
```diff
@@ -64,6 +65,11 @@ pub struct Header {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename = "x5t#S256")]
     pub x5t_s256: Option<String>,
+
+    /// Any additional headers
```
rimutaka, line 69: It would be good to give more info here - what it is, what it's for, what it looks like in serialized form.

src/header.rs
```diff
@@ -64,6 +65,11 @@ pub struct Header {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(flatten)]
+    pub extra: Option<HashMap<String, String>>,
```
rimutaka, line 72: Can the name be more self-explanatory? If we are to keep it short it should probably be `extras`.
I'm not sure what to call it. `extra` doesn't feel right.

tests/hmac.rs
```diff
@@ -41,7 +42,9 @@ fn encode_with_custom_header() {
         company: "ACME".to_string(),
         exp: OffsetDateTime::now_utc().unix_timestamp() + 10000,
     };
-    let header = Header { kid: Some("kid".to_string()), ..Default::default() };
+    let mut extra = HashMap::with_capacity(1);
```
rimutaka, line 45: I think we should have 3 tests:
- None
- 1 extra header
- 2 or more extra headers

src/header.rs
rimutaka, line 69: It would be good to mention this in the README as well.

rimutaka:
@jayakasadev , word of caution, I am not a maintainer, so feel free to ignore. The final word is with @Keats.

jayakasadev:
> @jayakasadev , word of caution, I am not a maintainer, so feel free to ignore. The final word is with @Keats.

understood. I wanted to ask if using a `trait` would be preferable. This way users who need a new field can define their own struct/impl without impacting existing users.

jayakasadev:
I went ahead and implemented the trait approach in a separate branch in case 
https://github.com/jayakasadev/jsonwebtoken/commit/32709ea0f33bb3fa20707cf0d95fcdf73fb4e58e

let me know if this approach is more desirable. The usage of `Hashmap` causes a small regression in performance. Traits does not.

rimutaka:
Immediate thoughts: it's a breaking change and not sure if the existing tests cover all the affected parts.

jayakasadev:
understood, im fine abandoning that approach if the performance regression is tolerable
```
     Running benches/jwt.rs (target/release/deps/jwt-10e1350ca031353b)
bench_encode            time:   [517.70 ns 520.80 ns 524.03 ns]
                        change: [+0.6197% +1.0974% +1.6803%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

bench_decode            time:   [919.45 ns 920.88 ns 922.50 ns]
                        change: [+4.8808% +5.1744% +5.4616%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
  ```
this is on macos arm64

File: src/jwk.rs [hunk 4]
```diff
@@ -116,7 +116,7 @@ impl<'de> Deserialize<'de> for KeyOperations {
         D: Deserializer<'de>,
     {
         struct KeyOperationsVisitor;
-        impl<'de> de::Visitor<'de> for KeyOperationsVisitor {
+        impl de::Visitor<'_> for KeyOperationsVisitor {
```
rimutaka, line 119: Why lifetime change?

jayakasadev: https://github.com/Keats/jsonwebtoken/actions/runs/13506827659/job/38042025098?pr=420

rimutaka: @jayakasadev, do you know if it's due to the new 1.85.0 elision rules?
I wonder if this change breaks pre-2024 editions.

jayakasadev: I can't say for sure. I don't have visibility into older runs on `1.84.0` or other `1.85.0` runs. If this came up for the first time, then I would hazard to guess that this is the case.

jayakasadev:
```
In Rust 2024, all in-scope generic parameters, including lifetime parameters, are implicitly captured when the use<..> bound is not present.
```
https://doc.rust-lang.org/edition-guide/rust-2024/rpit-lifetime-capture.html

jayakasadev: https://medium.com/@TechSavvyScribe/understanding-lifetime-elision-in-rust-5df771c532f7#:~:text=There%20are%20three%20lifetime%20elision%20rules%20that,lifetime%20is%20assigned%20to%20all%20output%20reference

rimutaka: The GH actions run the tests against the MSRV here https://github.com/Keats/jsonwebtoken/blob/master/.github/workflows/ci.yml#L40
Can you manually run that workflow in your fork to see if it works?

jayakasadev: like this?
https://github.com/jayakasadev/jsonwebtoken/pull/1

rimutaka:
@jayakasadev , I'm not a maintainer and I'm not sure how much Vincent trusts my reviews :)
I'll give it another good look in the next few days anyway. Most likely early next week.
Thank you for putting in good work. 🙏

src/crypto/mod.rs
```diff
@@ -74,8 +83,10 @@ pub fn verify(
     match algorithm {
         Algorithm::HS256 | Algorithm::HS384 | Algorithm::HS512 => {
             // we just re-sign the message with the key and compare if they are equal
-            let signed = sign(message, &EncodingKey::from_secret(key.as_bytes()), algorithm)?;
-            Ok(verify_slices_are_equal(signature.as_ref(), signed.as_ref()).is_ok())
```
Keats, line 78: We can't remove that for now. Can you keep the code in that PR strictly about the headers?

Keats (changes requested):

src/header.rs
```diff
@@ -64,6 +65,13 @@ pub struct Header {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename = "x5t#S256")]
     pub x5t_s256: Option<String>,
+
+    /// Any additional non-standard headers not defined in [RFC7515#4.1](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1).
+    /// Once serialized, all keys will be converted to fields at the root level of the header payload
+    /// Ex: Dict("custom" -> "header") will be converted to "{"typ": "JWT", ..., "custom": "header"}"
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(flatten)]
+    pub extras: Option<HashMap<String, String>>,
```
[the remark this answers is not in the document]
jayakasadev: good call. option is not needed

```
let mut extra = HashMap::with_capacity(1);
let header = Header { kid: Some("kid".to_string()), extras: extra.clone(), ..Default::default() };
println!("{}", serde_json::to_string(&header).unwrap());
```
the output is `{"typ":"JWT","alg":"HS256","kid":"kid"}`

src/header.rs
```diff
@@ -64,6 +65,12 @@ pub struct Header {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename = "x5t#S256")]
     pub x5t_s256: Option<String>,
+
+    /// Any additional non-standard headers not defined in [RFC7515#4.1](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1).
+    /// Once serialized, all keys will be converted to fields at the root level of the header payload
```
Keats, line 70: Do we want to expose that for serialization? Most of the cases I've seen are people needing to read values in the header I believe

jayakasadev: I'm not sure what you mean. Can you expand?

Keats: In my mind, the extras values were only usable for deserialization, eg you get a jwt with some random non compliant things in there but then I remember things like https://github.com/Keats/jsonwebtoken/issues/368#issuecomment-2305990342

rimutaka: Im not sure what Vincent means either. `flatten` works both ways: ser/deser. 🤔

Keats: I thought the extra fields were mostly used for JWTs coming from services and you would not need to actually expose serializing them at all to the user. In practice it seems some provider like Apple do expect you to add custom headers so maybe you do need to handle Serialize after all so no changes to that PR are needed I think

tests/hmac.rs
```diff
@@ -51,6 +52,56 @@ fn encode_with_custom_header() {
...
+        &token,
+        &DecodingKey::from_secret(b"secret"),
+        &Validation::new(Algorithm::HS256),
+    )
+    .unwrap();
+    assert_eq!(my_claims, token_data.claims);
+    assert_eq!("kid", token_data.header.kid.unwrap());
+    let extras = token_data.header.extras;
+    assert_eq!("header1", extras.get("custom1").unwrap().as_str());
+    assert_eq!("header2", extras.get("custom2").unwrap().as_str());
```
Keats, line 104: Can we add a decode test with some jwt with additional header?

Keats (approved):

jayakasadev:
@Keats thanks for approving
I don't have merge permissions. 
The blocking clippy tests are due to deprecated API.

Keats:
This is merged but since it's technically a breaking change, I'll wait a bit before releasing it

PR #421, opened 2025-03-05
Handle verification when multiple algorithms are allowed in the settings

jsdt:
Previously it was possible to allow multiple algorithms in the validation settings, but if multiple families were allowed, verifying would always fail (because it wanted the decoding key to match all of the families). This change supports settings multiple algorithms/families. This will return an error if:
 1. The decoding key doesn't match the family of any of the allowed algorithms,
 2. The algorithm in the token header is not one of the allowed algorithms, or
 3. The algorithm in the token header is not the same family as the decoding key

I added some tests for these cases. I also added `decode_token_with_multiple_algorithms_allowed`, which would have previously failed.

### DIFF
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -213,10 +213,10 @@ fn verify_signature<'a>(
         return Err(new_error(ErrorKind::MissingAlgorithm));
     }
 
-    if validation.validate_signature {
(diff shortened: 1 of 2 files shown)


Keats:
The early error with mixed family is intended. Ideally it would not be possible at all from the Rust API but I haven't got to that yet

PR #428, opened 2025-05-11
Decoupled crypto backends (the rest)

sulami:
This is following up on #410 with the rest of the decoupled implementations, and addressing most of the review comments (I think?).

There are still some todos left for documentation, which I haven't filled in yet.

I've verified that tests & clippy pass with all combinations of features, but a lot of this I only know enough about to be dangerous, so I'd appreciate close looks at the actual implementations to make sure e.g. I've picked the correct library functions in each case.


Keats:
:o how did i miss this sorry, i'll go through it asap

Cargo.toml
```diff
@@ -48,8 +60,10 @@ time = { version = "0.3", features = ["wasm-bindgen"] }
 criterion = { version = "0.4", default-features = false }
 
 [features]
-default = ["use_pem"]
+default = ["use_pem", "aws_lc_rs"]
```
Keats, line 63: should we have a default crypto backend or let users pick the one they want?

src/crypto/aws_lc/ecdsa.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        Ok(())
+    }
+}
+
+impl JwtVerifier for Es384Verifier {
+    fn algorithm(&self) -> Algorithm {
+        Algorithm::ES384
+    }
+}
```
Keats, line 126: Can those be abstracted with a macro? It's the same stuff except for a few elemens

src/crypto/aws_lc/hmac.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn verify(&self, msg: &[u8], signature: &Vec<u8>) -> std::result::Result<(), signature::Error> {
+        hmac::verify(&self.0, msg, signature).map_err(signature::Error::from_source)
+    }
+}
+
+impl JwtVerifier for Hs512Verifier {
+    fn algorithm(&self) -> Algorithm {
+        Algorithm::HS512
+    }
+}
```
Keats, line 150: same, can it be macroed?

src/crypto/aws_lc/rsa.rs
```diff
@@ -0,0 +1,340 @@
...
+    fn verify(&self, msg: &[u8], signature: &Vec<u8>) -> std::result::Result<(), signature::Error> {
+        verify_rsa(&crypto_sig::RSA_PSS_2048_8192_SHA512, &self.0, msg, signature)
+    }
+}
+
+impl JwtVerifier for RsaPss512Verifier {
+    fn algorithm(&self) -> Algorithm {
+        Algorithm::PS512
+    }
+}
```
Keats, line 340: same question for macros

src/crypto/rust_crypto/hmac.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        verifier.verify_slice(signature).map_err(signature::Error::from_source)
+    }
+}
+
+impl JwtVerifier for Hs512Verifier {
+    fn algorithm(&self) -> Algorithm {
+        Algorithm::HS512
+    }
+}
```
Keats, line 185: same question for macros for all the rust_crypto files as well.

src/crypto/utils.rs
```diff
@@ -0,0 +1,25 @@
...
+//!
+//! - Put in documentation
+
+use crate::{
+    algorithms::AlgorithmFamily,
+    errors::{new_error, ErrorKind, Result},
+    DecodingKey, EncodingKey,
+};
+
+pub(crate) fn try_get_hmac_secret_from_encoding_key(encoding_key: &EncodingKey) -> Result<&[u8]> {
```
Keats, line 11: should this be on EncodingKey directly?

src/crypto/utils.rs
```diff
@@ -0,0 +1,25 @@
...
+    if encoding_key.family == AlgorithmFamily::Hmac {
+        Ok(encoding_key.inner())
+    } else {
+        Err(new_error(ErrorKind::InvalidKeyFormat))
+    }
+}
+
+pub(crate) fn try_get_hmac_secret_from_decoding_key(decoding_key: &DecodingKey) -> Result<&[u8]> {
```
Keats, line 19: and this in DecodingKey?

src/decoding.rs
```diff
@@ -286,3 +312,41 @@ pub fn decode_header(token: &str) -> Result<Header> {
...
+fn verify_signature<'a>(
+    token: &'a str,
+    validation: &Validation,
+    verifying_provider: Box<dyn JwtVerifier>,
+) -> Result<(Header, &'a str)> {
+    if validation.validate_signature && validation.algorithms.is_empty() {
+        return Err(new_error(ErrorKind::MissingAlgorithm));
+    }
+
+    // Todo: This behaviour is currently not captured anywhere.
```
Keats, line 328: If you have an idea on how to improve that, i'll take it

tests/hmac.rs
```diff
@@ -1,38 +1,46 @@
...
-#[test]
-#[wasm_bindgen_test]
-fn sign_hs256() {
-    let result =
-        sign(b"hello world", &EncodingKey::from_secret(b"secret"), Algorithm::HS256).unwrap();
-    let expected = "c0zGLzKEFWj0VxWuufTXiRMk5tlI5MbGDAYhzaxIYjo";
-    assert_eq!(result, expected);
-}
+// #[test]
+// #[wasm_bindgen_test]
```
Keats, line 28: can we re-enable the wasm tests? are they failing with aws_lc?

Keats (changes requested):
Looks great overall. I think some macros would help reduce the duplication

Cargo.toml
sulami: I mean, that's up to you. I personally think it's nice to have a reasonable default.

sulami:
I'll have a go at reducing the duplication through some macros and addressing the other comments.

tests/hmac.rs
sulami: Ah no, the `sign` and `verify` functions were gone when I picked up the partial PR, but I've re-added them and it works just fine.

src/decoding.rs
sulami: This comment is actually @sidrubs', I'm not entirely sure what they mean by that, whether it's not covered by tests, or docs, or...?

sulami:
Alright, done some refactoring to reduce the duplication and re-enabled the WASM HMAC test cases. I checked that tests pass with both AWS-LC and Rust-crypto.

src/decoding.rs
Keats: We probably want to uncomment it for now, but ideally we find a better way to handle it before releasing the next version since it trips up people

tests/rsa/mod.rs
```diff
@@ -26,6 +26,8 @@ pub struct Claims {
     exp: i64,
 }
 
+// Todo: These no longer apply because `verify` does not exist, would probably need to convert it to test the factory for getting signers and verifiers. But I would rather this not be part of the public facing API.
```
Keats, line 29: is it still a todo?

Keats (commented):
That looks great. I'll have another look before merging but it's a great improvement!

tests/rsa/mod.rs
sulami: Ah true, missed that one. `verify` does indeed exist again.

sulami:
Addressed the latest comments, and also inlined `_encode` and `_decode`, which were single use, for some reason `pub`, and had `TODO` as docstrings.

sulami:
@Keats Just in case you missed it again, I think this should be good for another review now.

src/crypto/rust_crypto/rsa.rs
```diff
@@ -0,0 +1,145 @@
+//! Implementations of the [`JwtSigner`] and [`JwtVerifier`] traits for the
+//! RSA family of algorithms using [`aws_lc_rs`]
```
Keats, line 2:
```suggestion
//! RSA family of algorithms using [`RustCrypto`]
```

Keats (approved):
That looks good. One small copy/paste issue but fine otherwise. I need to create a 0.10 branch to group everything

src/crypto/rust_crypto/rsa.rs
sulami: Ah, good catch, missed that. Pushed a fix.

sulami:
Looking at [the AWS-LC docs](https://github.com/aws/aws-lc#supported-platforms), I don't think it builds/is supported on WASM. I've also recently read [this](https://github.com/launchbadge/sqlx/tree/main/examples/postgres/preferred-crates) with regards to default features, and I'm thinking maybe we shouldn't have a default backend after all? That way we could also pick a different backend for the WASM build.

Keats:
Yeah maybe let's not have a default if the default doesn't work with wasm.

sulami:
I don't know if this'll work or not, at least it builds on my machine, but it doesn't pass the WASM tests. Neither does master though, so I'm not sure that counts for much.

I'm not terribly familiar with WASM overall, so I wouldn't mind additional pointers & opinions.

Cargo.toml
```diff
@@ -43,6 +43,8 @@ sha2 = { version = "0.10.7", optional = true, features = ["oid"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 js-sys = "0.3"
+# https://docs.rs/getrandom/0.2.16/getrandom/index.html#webassembly-support
+getrandom = { version = "0.2.16", features = ["js"] }
```
jplatte, line 47: > We only enable it if compiling for WASM, following getrandom's advice to not enable it unconditionally for our users.

Their docs (also) say this though:

> This feature should only be enabled for binary, test, or benchmark crates. Library crates should generally not enable this feature, leaving such a decision to users of their library.

I think this commit should be reverted, and CI adjusted to run WASM checks / tests with `--features getrandom/js`.

.github/workflows/ci.yml
```diff
@@ -80,7 +80,7 @@ jobs:
                 run: cargo install wasm-pack
 
             -   name: Run tests default features
-                run: wasm-pack test --node --features rust_crypto
+                run: wasm-pack test --node --features rust_crypto,getrandom/js
```
Keats, line 83: There's a related issue: https://github.com/Keats/jsonwebtoken/issues/431
But I don't know enough about wasm to know what to actually do

jplatte: I think this PR might actually help with that, given `getrandom/js` is now only enabled in CI. The user may be able to enable another wasm-compatible `getrandom` backend after this lands.

Keats:
Any ideas why the jobs are stuck?

.github/workflows/ci.yml
```diff
@@ -1,83 +1,86 @@
...
+    tests:
+        name: Tests
+        runs-on: ${{ matrix.os }}
+        strategy:
+            matrix:
+                build: [ pinned, stable, nightly ]
+                backend: [ aws_lc_rs, rust_crypto ]
+                include:
+                    -   build: pinned
+                        os: ubuntu-20.04
```
jplatte, line 43: I believe ubuntu-20.04 is no longer supported by GHA. Need tp upgrade to 22.04 or 24.04.

Keats:
looks like we need to bump the pinned version in CI as well

Keats:
All good! I'll probably release a new version based on this PR this week if I don't forget

Issue #438, opened 2025-10-01
`DecodingKey` is validated even when disabling signature validation


bombsimon:
Thanks for this module, really useful!

I've been using it to decode JWT tokens without verifying them because that's not important and any concern in this specific workflow. Up until `v10.0.0` I've been able to pass any `DecodingKey` to any validation signature as long as signature validation has been turned off, something like this:

```rust
#[derive(Clone, Debug, serde::Deserialize)]
#[serde(rename_all = "camelCase")]
struct Claims {
    some_claim: String,
}

// Setting RS265 here to match the JWT
let mut validation = jsonwebtoken::Validation::new(jsonwebtoken::Algorithm::RS256);
validation.validate_aud = false;
validation.insecure_disable_signature_validation(); // Disabling validation
                                                                           
// Easiest to create `DecodingKey` but this does _not_ match the signature (this is Hmac)          
let dummy_key = jsonwebtoken::DecodingKey::from_secret(&[]);

let decoded = jsonwebtoken::decode::<Claims>(token, &dummy_key, &validation)
    .expect("failed to read claims: {err}"))?;
```

Trying to run this code after `v10.0.0` now results in `InvalidKeyFormat` since even though I'm not interested in verifying the signature we call `jwt_verifier_factory` which will create a verifier based on the header and pass the key which may contain a different algorithm.

One fix for this is to simply pass a dummy RSA key instead, either a static one generated with something like `openssl genrsa 512 | openssl rsa -pubout`, or pull in a dependency on `openssl` and use `openssl::rsa::Rsa::generate()`.

Unless there is some other more idiomatic way to do this? Reading the pinned issue, my interpretation is that this library is strongly against not verifying tokens so maybe this is a very intentional decision and one step further away from simplifying decoding without verifying?

Thanks!

bombsimon:
Thanks for a swift reply! 

Do you want me to look into it and PR a change? Maybe we could conditionally call `jwt_verifier_factory` and make `verifying_provider` optional in `verify_signature`, then we return the payload (claims) as is if `validation.validate_signature` is `false` and `jwt_verifier_factory` is `None`?

dsykes16:
This is blocking `spiffe` from updating as well. Implementing #401 would meet the requirements for SPIFFE.

PR #441, opened 2025-10-07
feat: add `dangerous::insecure_decode`

dsykes16:
Add `jsonwebtoken::dangerous::insecure_decode` to support decoding headers and claims with no signature validation.

This should fulfill #401 and also provides a solution for #438

### DIFF
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -292,6 +292,23 @@ pub fn decode<T: DeserializeOwned + Clone>(
     Ok(TokenData { header, claims })
 }
 
+/// Decode a JWT with NO VALIDATION
+///
+/// DANGER: This performs zero validation on the JWT
+pub fn insecure_decode<T: DeserializeOwned + Clone>(
+    token: impl AsRef<[u8]>,
+) -> Result<TokenData<T>> {
+    let token = token.as_ref();
+
+    let (_, message) = expect_two!(token.rsplitn(2, |b| *b == b'.'));
+    let (payload, header) = expect_two!(message.rsplitn(2, |b| *b == b'.'));
+
+    let header = Header::from_encoded(header)?;
+    let claims = DecodedJwtPartClaims::from_jwt_part_claims(payload)?.deserialize()?;
+
+    Ok(TokenData { header, claims })
+}
+
 /// Return the correct [`JwtVerifier`] based on the `algorithm`.
 pub fn jwt_verifier_factory(
     algorithm: &Algorithm,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,11 @@ pub use encoding::{EncodingKey, encode};
 pub use header::Header;
 pub use validation::{Validation, get_current_timestamp};
 
+/// Dangerous decoding functions that should be audited and used with extreme care.
+pub mod dangerous {
+    pub use super::decoding::insecure_decode;
+}
+
 mod algorithms;
 /// Lower level functions, if you want to do something other than JWTs
 pub mod crypto;
--- a/tests/dangerous.rs
+++ b/tests/dangerous.rs
@@ -1,0 +1,60 @@
+use jsonwebtoken::{Algorithm, TokenData, dangerous::insecure_decode};
+use wasm_bindgen_test::wasm_bindgen_test;
+
+#[derive(Debug, PartialEq, Eq, Clone, serde::Serialize, serde::Deserialize)]
+pub struct Claims {
+    sub: String,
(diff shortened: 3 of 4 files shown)


src/decoding.rs
```diff
@@ -292,6 +292,23 @@ pub fn decode<T: DeserializeOwned + Clone>(
     Ok(TokenData { header, claims })
 }
 
+/// Decode a JWT with NO VALIDATION
+///
+/// DANGER: This performs zero validation on the JWT
+pub fn insecure_decode<T: DeserializeOwned + Clone>(
```
Keats, line 298: What should we do with https://docs.rs/jsonwebtoken/9.3.1/jsonwebtoken/struct.Validation.html#method.insecure_disable_signature_validation? Deprecate it? Is there a place where you want to validate a JWT claims but not the signature? That sounds useless

Keats (commented):
LGTM

src/decoding.rs
dsykes16: I'll defer to you and other maintainers on that. This new `jsonwebtoken::dangerous::insecure_decode` fulfill's SPIFFE's usecase exactly, and probably works for any other folks that need to take a peek at the claims to decide which key to use for validation.

The existing `insecure_disable_signature_validation` is still somewhat backwards compatible between v10 and v9 as long as the algorithm in the JWT header matches the placeholder key passed in the `Verifier`. I'm not sure how common that pattern is. I'm guessing most folks using the existing `insecure_disable_signature_validation` path are passing in `DecodingKey::from_secret(&[])`, but in my experience `HMAC` isn't actually that common in JWT's these days.

dsykes16: Maybe an option is marking `insecure_disable_signature_validation` as planned for deprecation in v11, and leveraging the `insecure_decode` internally if `validation.validate_signature == false`.

```
diff --git a/src/decoding.rs b/src/decoding.rs
index 7ebb8fd..1a1db36 100644
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -275,9 +275,14 @@ pub fn decode<T: DeserializeOwned + Clone>(
     validation: &Validation,
 ) -> Result<TokenData<T>> {
     let token = token.as_ref();
+
+    if !validation.validate_signature {
+        return insecure_decode::<T>(token);
+    }
+
     let header = decode_header(token)?;

-    if validation.validate_signature && !validation.algorithms.contains(&header.alg) {
+    if !validation.algorithms.contains(&header.alg) {
         return Err(new_error(ErrorKind::InvalidAlgorithm));
     }

@@ -356,25 +361,21 @@ pub(crate) fn verify_signature_body(
     validation: &Validation,
     verifying_provider: Box<dyn JwtVerifier>,
 ) -> Result<()> {
-    if validation.validate_signature && validation.algorithms.is_empty() {
+    if validation.algorithms.is_empty() {
         return Err(new_error(ErrorKind::MissingAlgorithm));
     }

-    if validation.validate_signature {
-        for alg in &validation.algorithms {
-            if verifying_provider.algorithm().family() != alg.family() {
-                return Err(new_error(ErrorKind::InvalidAlgorithm));
-            }
+    for alg in &validation.algorithms {
+        if verifying_provider.algorithm().family() != alg.family() {
+            return Err(new_error(ErrorKind::InvalidAlgorithm));
         }
     }

-    if validation.validate_signature && !validation.algorithms.contains(&header.alg) {
+    if !validation.algorithms.contains(&header.alg) {
         return Err(new_error(ErrorKind::InvalidAlgorithm));
     }

-    if validation.validate_signature
-        && verifying_provider.verify(message, &b64_decode(signature)?).is_err()
-    {
+    if verifying_provider.verify(message, &b64_decode(signature)?).is_err() {
         return Err(new_error(ErrorKind::InvalidSignature));
     }
```

That would fully resolve #438 by keeping backwards compatibility with `insecure_disable_signature_validation` although it does appear to change existing behavior in that it doesn't check the `exp` anymore (expected by `tests::hmac::dangerous_insecure_decode_token_with_validation_wrong_algorithm`). It also cleans up a lot of the conditional statements in the verification path.

dsykes16: @Keats, what are you thoughts on this? I suppose it all comes down to how strict we want to be with SemVer since v10 is already out, albeit with no docs on docs.rs yet.

Keats: let's just deprecate it for now and we will remove it for v11

Keats (approved):

mokhaled2992:
Any idea when will this get released?

PR #444, opened 2025-10-09
Custom header support

### LINKED ISSUE #439: Header doesn't implement `Hash` anymore
Hey, thanks for all your work maintaining jsonwebtoken.

It seems  that in the change to allow swapping crypto backends that [hash](https://github.com/Keats/jsonwebtoken/commit/5cd1887848f7d30f11adb4ebcec6a01baffd5252#diff-5991004fd31cd8d58f2dbd269886018796101de043c3f0c8897a03ab049054a5L13-L65) was removed from header. This is kind of repeat of the issue that Craig reported in #237.

Was this an oversight?

Thanks!

dsykes16:
In service of #439 

Adds support for custom headers via `decode_with_custom_header` to decode and validate JWT, and a standalone header-only decode via `decode_custom_header`. Those should probably have better names before this is merged. The existing `decode` and `decode_header` functions retain full backwards compatibility.

Also adds convenience macros for claims and custom headers, `jsonwebtoken::macros::claims` and `jsonwebtoken::macros::header`, and a `Header` derive macro.

Adds a `BasicHeader` to fulfill #439 (i.e. implement `Hash`). I have mixed feelings on that one. I think ideally `extras` wouldn't exist on the original `Header` struct, which was itself a breaking change (#420) due to removing the `Hash` implementation, but removing it now would be a breaking change. Given those constraints, adding a `BasicHeader` is the only path I could think of to resolve #439 and also to avoid the minor performance hit of using a HashMap (along w/ the inherent risk of a malicious user intentionally sending JWTs with abnormally large headers containing a large number fields).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,9 @@ rand = { version = "0.8.5", optional = true, features = ["std"], default-feature
 rsa = { version = "0.9.6", optional = true }
 sha2 = { version = "0.10.7", optional = true, features = ["oid"] }
 
+# proc-macros (e.g. Header, claims)
+jsonwebtoken-proc-macros = { path = "jsonwebtoken-proc-macros" }
+
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 js-sys = "0.3"
 getrandom = "0.2"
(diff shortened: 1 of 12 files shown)


arckoor:
Done in #512

PR #445, opened 2025-10-09
decoding: add `decode_header_as` function

### LINKED ISSUE #435: Support for decoding with issuer-based decode keys
Would it be reasonable to support decoding using a `lookup_key(&str) -> Option<&DecodingKey>` function? This would allow the decoding to not have to create a validation per known key with varying issuer validation.

Or perhaps this is way off target with how JWTs are meant to be used? But it feels weird to have to "guess" at which decoding key to use if it is based on the `iss` field.

I took a peek at the code and it would require a bit of surgery and "stuttering" of decoding to extract out just the `iss` field before moving onto validation.

mathstuf:
Headers may contain custom fields. As an example, these fields may be used by implementations to choose the decoding key rather than disabling signature validation, parsing the payload, determining the key and *then* decoding securely.

RFC 7519 § 5.3 explicitly calls out the ability to store `iss`, `sub`, and `aud` in the header.

Using custom structures also allows users to avoid storing any unnecessary data when doing such pre-processing.

See: https://datatracker.ietf.org/doc/html/rfc7519#section-5.3
Fixes: #435

### DIFF
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -357,6 +357,33 @@ pub fn decode_header(token: impl AsRef<[u8]>) -> Result<Header> {
     Header::from_encoded(header)
 }
 
+/// Decode a JWT without any signature verification/validations and return its header as a custom type.
+///
+/// If the token has an invalid format (ie 3 parts separated by a `.`), it will return an error.
+///
+/// ```rust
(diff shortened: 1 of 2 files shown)


arckoor:
#512 should provide everything this seeks to accomplish

PR #447, opened 2025-10-10
Add botan as a crypto backend

arckoor:
Uses [botan-rs](https://github.com/randombit/botan-rs) (which wrapps the FFI of [botan](https://github.com/randombit/botan)) to provide another crypto backend.

Some notes:
- as an FFI wrapper, botan-rs *can* emit a lof of errors (mainly related to passing the wrong stuff through the interface), which is why there's a lot of `?`s in the code. In practice almost none of these will ever happen, but `?` feels cleaner to methan `unwrap()`. The only place that needs to unwrap is `compute_digest`.
- this currently uses my forks of botan(-rs) (hence the draft), because I needed to add the ability to load a PKCS1 RSA public key. I'm reasonably sure I can get this PR'd into botan, but I'd like to make sure this PR can have success before I go adding new features to something as old as RSA.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                backend: [ aws_lc_rs, rust_crypto ]
+                backend: [ aws_lc_rs, botan, rust_crypto ]
         steps:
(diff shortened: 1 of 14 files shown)


arckoor:
That would be appreciated too, I'll close this for the time being then.
Feel free to ping me if I can help in some way getting this done.

PR #452, opened 2025-10-17
Expose cryptography backends via CryptoProvider

arckoor:
Adds a `CryptoProvider` struct that allows replacing the built-in providers with something custom.
All the details from this implementation that could be considered "interesting" are stolen straight from `rustls`'s [CryptoProvider](https://github.com/rustls/rustls/blob/706d1600248a982ebbe894c0b71738b115b58e4b/rustls/src/crypto/mod.rs).

I've marked the `new_signer`, `new_verifier` and JWK functions from the two built in backends as `pub`, so you can do stuff like this:
```rs
fn new_signer(algorithm: &Algorithm, key: &EncodingKey) -> Result<Box<dyn JwtSigner>, Error> {
    let jwt_signer = match algorithm {
        Algorithm::EdDSA => Box::new(CustomEdDSASigner::new(key)?) as Box<dyn JwtSigner>,
        _ => jsonwebtoken::crypto::aws_lc::new_signer(algorithm, key)?,
    };

    Ok(jwt_signer)
}
```
i.e. overwrite just specific algorithms.

One area I'm a little unsure about is `JwkUtils`, 1) about the name and 2) about the `Default` implementation. The `CryptoProvider::signer_` and `CryptoProvider::verifier_factory` functions are obviously mandatory for a custom provider, but not everyone uses JWK, so the default just uses dummy functions with `unimplemented!()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,10 @@ jsonwebtoken = { version = "10", features = ["aws_lc_rs"] }
 serde = {version = "1.0", features = ["derive"] }
 ```
 
-Two crypto backends are available via features, `aws_lc_rs` and `rust_crypto`, exactly one of which must be enabled.
+Two crypto backends are available via features, `aws_lc_rs` and `rust_crypto`, at most one of which must be enabled. If you select neither feature, you need to provide your own `CryptoProvider`.
+
+For examples of how to implement a `CryptoProvider`, see
+- [arckoor/jsonwebtoken-botan](https://github.com/arckoor/jsonwebtoken-botan)
 
 The minimum required Rust version (MSRV) is specified in the `rust-version` field in this project's [Cargo.toml](Cargo.toml).
 
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -1,20 +1,26 @@
 //! The cryptography of the `jsonwebtoken` crate is decoupled behind
-//! [`JwtSigner`] and [`JwtVerifier`] traits. These make use of `RustCrypto`'s
+//! [`JwtSigner`] and [`JwtVerifier`] traits. These make use of `signature`'s
 //! [`Signer`] and [`Verifier`] traits respectively.
+//! Crypto provider selection is handled by [`CryptoProvider`].
 //!
 //! [`JwtSigner`]: crate::crypto::JwtSigner
 //! [`JwtVerifier`]: crate::crypto::JwtVerifier
 //! [`Signer`]: signature::Signer
 //! [`Verifier`]: signature::Verifier
+//! [`CryptoProvider`]: crate::crypto::CryptoProvider
 
 use crate::algorithms::Algorithm;
 use crate::errors::Result;
+use crate::jwk::{EllipticCurve, ThumbprintHash};
 use crate::{DecodingKey, EncodingKey};
 
+/// `aws_lc_rs` based CryptoProvider.
 #[cfg(feature = "aws_lc_rs")]
-pub(crate) mod aws_lc;
+pub mod aws_lc;
+
+/// `RustCrypto` based CryptoProvider.
 #[cfg(feature = "rust_crypto")]
-pub(crate) mod rust_crypto;
+pub mod rust_crypto;
 
 use crate::serialization::{b64_decode, b64_encode};
 use signature::{Signer, Verifier};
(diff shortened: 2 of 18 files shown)


arckoor:
@Keats any chance of a review on this?

src/crypto/mod.rs
```diff
@@ -58,6 +71,130 @@ pub fn verify(
...
+    pub(crate) fn install_default(
+        default_provider: CryptoProvider,
+    ) -> Result<(), Arc<CryptoProvider>> {
+        PROCESS_DEFAULT_PROVIDER.set(Arc::new(default_provider))
+    }
+
+    pub(crate) fn get_default() -> Option<&'static Arc<CryptoProvider>> {
+        PROCESS_DEFAULT_PROVIDER.get()
+    }
+}
```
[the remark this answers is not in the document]
DoumanAsh: This should be feature of power user, not regular user.
Most people prefer to add `jsonbwebtoken` and it just works
It is bug that rustls require you to install specific backend when both aws-lc and ring are enabled as features

The best user experience if we define following logic:
- If user installed provider, return it
- Otherwise return following in order of features combo:
     - If both rust-crypto and aws-lc defined then return awc-lc
     - If aws-lc defined return aws-lc
     - If rust-crypto defined then return rust-cryptop

DoumanAsh (commented):
I left some comments how to implement it better and simplify provider logic
I think it would be also great to have NOOP provider implementation to be used by default instead of relying on `Option` to reduce indirection

arckoor:
@DoumanAsh I can sympathize with most of the comments, and will try them out soon-ish. What I don't sympathize with is returning a default provider when both features are enabled. If there is a default provider, it should be a default feature. Making cryptographic choices for the user that they might not know about is imo a very bad idea.

Keats:
I agree with @arckoor if several features are enabled in the tree but the user didn't pick. That's a one time choice for the user but for example aws-ls-rc doesn't work with WASM afaik so I would rather not pick the backend library without a clear input in case of conflict.

DoumanAsh:
@Keats The problem is that feature can be enabled by your dependency, it can be alleviated  by having NOOP backend so that library could encourage others not to enable features I guess, but if for some reason you accidentally enable both features, you have no way to diagnose it simply, so that's my concern

Keats:
Sorry for the delay. Let's point to other backends for examples, no need to maintain one ourselves.

README.md
```diff
@@ -15,7 +15,8 @@ jsonwebtoken = { version = "10", features = ["aws_lc_rs"] }
 serde = {version = "1.0", features = ["derive"] }
 ```
 
-Two crypto backends are available via features, `aws_lc_rs` and `rust_crypto`, exactly one of which must be enabled.
+Two crypto backends are available via features, `aws_lc_rs` and `rust_crypto`, at most one of which must be enabled. If you select neither feature, you need to provide your own `CryptoProvider`.
+See e.g. [arckoor/jsonwebtoken-botan](https://github.com/arckoor/jsonwebtoken-botan) for an example of how to implement a `CryptoProvider`.
```
Keats: let's turn that into a list linking to the various providers

DoumanAsh:
Functionally wise it solves issue, but generally I think Arc and Option are adding unnecessary indirection to access provider (I pointed it out in my review but as I'm not really owner I didn't insist on it)
If you think it is ok, then I'm fine with it too (alternative approach is to use `'static` reference to provider and store pointer to it, similarly how `log` crate does with logger)

Keats:
> So that is completely a case of "you decide @Keats".

:sweat_smile: I really do not care personally but if Arc is not needed in any way we can remove it

DoumanAsh:
@arckoor @Keats Here is quick approach to remove Arc and Option
https://github.com/DoumanAsh/jsonwebtoken/commit/5edada7f9cfa81504726b3805febbc075b4a26bb

It moves everything into static variables and relies on `OnceLock` for initialisation to benefit from its optimisations instead of manually checking Option/install default on first access

src/crypto/mod.rs
```diff
@@ -58,6 +64,124 @@ pub fn verify(
...
+        }
+
+        #[cfg(all(feature = "aws_lc_rs", not(feature = "rust_crypto")))]
+        {
+            return &aws_lc::DEFAULT_PROVIDER;
+        }
+
+        #[allow(unreachable_code)]
+        {
+            const NOT_INSTALLED_ERROR: &str = r###"
```
DoumanAsh: This constant is duplicate of `NOT_INSTALLED_OR_UNIMPLEMENTED_ERROR`
Should move it to the top of module and re-use it?

DoumanAsh:
I believe `non_exhaustive` would be considered breaking change as it will affect direct use of `Algorithm` enum when trying to match it
It is probably best to add this attribute after release of the crypto provider to ensure current version is not broken

Keats:
Ah yes, let's remove just that bit then, we can add it later

Keats:
Looks like it could go out this week in a 10.3.0 release?

PR #456, opened 2025-10-20
Improve error messages for feature flags

omid:
Basically "At least one..." means we can enable both


arckoor:
The compile checks don't exist anymore since with #452 any number of providers can be enabled simultaneously

Issue #432, opened 2025-07-12
Consider to add Clone trait Bound to decode


dancixx:
Please consider adding `+ Clone` in the `decode` function. I am getting more and more often tait bound error in this decode function when I try to use this with generics.

```
error[E0277]: the trait bound `T: Clone` is not satisfied
   --> /Users/danixx/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/jsonwebtoken-9.3.1/src/decoding.rs:281:13
    |
281 |             Ok(TokenData { header, claims })
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `T`
    |
note: required by a bound in `TokenData`
   --> /Users/danixx/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/jsonwebtoken-9.3.1/src/decoding.rs:18:8
    |
16  | pub struct TokenData<T>
    |            --------- required by a bound in this struct
17  | where
18  |     T: Clone,
    |        ^^^^^ required by this bound in `TokenData`
help: consider further restricting type parameter `T` with trait `Clone`
    |
269 | pub fn decode<T: DeserializeOwned + std::clone::Clone>(
    |             
```

Possibly solution:

```rust
pub fn decode<T: DeserializeOwned + Clone>(
    token: &str,
    key: &DecodingKey,
    validation: &Validation,
) -> Result<TokenData<T>> {
    match verify_signature(token, key, validation) {
        Err(e) => Err(e),
        Ok((header, claims)) => {
            let decoded_claims = DecodedJwtPartClaims::from_jwt_part_claims(claims)?;
            let claims = decoded_claims.deserialize()?;
            validate(decoded_claims.deserialize()?, validation)?;

            Ok(TokenData { header, claims })
        }
    }
}
```

PR #458, opened 2025-10-29
Remove unnecessary Clone bounds from decode functions

okkero:
See discussion on https://github.com/Keats/jsonwebtoken/issues/432


LukasKalbertodt:
Thanks a ton :)

PR #459, opened 2025-11-09
Feature Discussion: Make rsa optional

kstiehl:
# Description

This PR should serve as the basis of a discussion since I am not sure myself if it is a good idea or not. But maybe a few words on how I got here. I am currently using this crate in one of my projects and I really like it so far. However the audit tools complain about [this vulnerability (RUSTSEC-2023-0071)](https://rustsec.org/advisories/RUSTSEC-2023-0071). 
I am not 100% sure if this vulnerability affects the `jsonwebtoken` crate or not and its also not that important since I am using no RSA functionality. 

This brought me to the code  and it looked like all the `rsa` parts can be made optional and it is still a very useful crate. 
To not break everything for everyone it should remain a `default` feature for now so users of this library would not have a breaking change, but people who want to skip `rsa` can do so.

## Questions for Discussion

1. Does this approach make sense for the project?
2. Should we also make RSA optional when using `aws_lc_rs`?
3. Are there any concerns about feature flag complexity?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,10 +64,10 @@ time = { version = "0.3", features = ["wasm-bindgen"] }
 criterion = { version = "0.4", default-features = false }
 
 [features]
-default = ["use_pem"]
+default = ["use_pem", "rsa"]
 use_pem = ["pem", "simple_asn1"]
-rust_crypto = ["ed25519-dalek", "hmac", "p256", "p384", "rand", "rsa", "sha2"]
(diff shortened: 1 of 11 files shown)


Keats:
1. Not really I think? You have the choice of multiple backends and people can just deny that warning if they want to use rust-crypto and don't use RSA
2. No reason to
3. Yeah, it's a feature flag for a feature flag

Issue #297, opened 2023-03-10
Invalid multiple algorithm logic


ereOn:
The `Validation` takes a `Vec` of `Algorithm` but the validation code [here](https://github.com/Keats/jsonwebtoken/blob/147d24170497a5aa973c74ac4566e8baab97770c/src/decoding.rs#L201-205) incorrectly ensures that the JWT's algorithm is compatible with **all** the specified algorithm, instead of just ensuring it finds at least one that matches.

As it stands, it is impossible to use more than one algorithm in a validation because of this bug.

Keats:
> As it stands, it is impossible to use more than one algorithm in a validation because of this bug.

Yeah that's intentional. Ideally we would solve it at the type level to only allow a single family but I didn't want to change the types too much.

Keats:
Most of the time, the `Validation` struct can be defined in a const way (well lazy_static or once_cell).

> Is it possible the Validation struct gained a responsibility to ensure a specific DecodingKey is appropriate rather than the DecodingKey's family being used for this?

As in passing the decoding key to the validation? It's possible but you lose the build once only approach

PeteX:
Yes I've got the same use case as @beanow-at-crabnebula. If the current behaviour is intentional, can I suggest updating the documentation? It currently implies that it's enough for the token to match a single algorithm from the list:

> The validation will check that the `alg` of the header is contained in the ones provided and will error otherwise.

Keats:
Honestly IMO we probably shouldn't even have the algorithm in the Validation at all since we have a DecodingKey which has the AlgorithmFamily baked in. You might still want to enforce a specific algorithm in a family though.

ytimenkov:
> The original bug was that you can construct a Validation struct that never passes validation, by mixing families. In my opinion a very common use-case when you're using JWKS to dynamically load in public keys with an allowlist of algorithms.

Exactly my use-case - for my dev environment I want to use public asymmetric keys from OIDC and a symmetric key. This just can't possibly pass because one family is RSA and another is HMAC.

I get that you don't want to allow `none` algorithm passed to 3rd-party validator, but if I found proper key by id I see no reason why this extra check is needed.

arckoor:
`Validation` may be redesigned in the future, see #520 for some discussion on it

p-lindberg:
[...]

PR #466, opened 2025-11-18
algorithms: expose `AlgorithmFamily`

mathstuf:
It is available on `DecodingKey`. Since `Validation` ends up masking out keys based on a single family, allow crates to create a list of supported algorithms based on the family in use.

See: #297

### DIFF
--- a/src/validation.rs
+++ b/src/validation.rs
@@ -6,7 +6,7 @@ use std::marker::PhantomData;
 use serde::de::{self, Visitor};
 use serde::{Deserialize, Deserializer};
 
-use crate::algorithms::Algorithm;
+use crate::algorithms::{Algorithm, AlgorithmFamily};
 use crate::errors::{ErrorKind, Result, new_error};
 
 /// Contains the various validations that are applied after decoding a JWT.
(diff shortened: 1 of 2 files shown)


Keats (approved):

Keats (approved):

arckoor:
@mathstuf can you rebase?

src/validation.rs
```diff
@@ -111,12 +111,21 @@ pub struct Validation {
 impl Validation {
     /// Create a default validation setup allowing the given alg
     pub fn new(alg: Algorithm) -> Validation {
+        Self::new_impl(vec![alg])
+    }
+
+    /// Create a default validation setup allowing any algorithm in the family
+    pub fn new_for_family(family: AlgorithmFamily) -> Validation {
+        Self::new_impl(family.algorithms().into_iter().copied().collect())
```
arckoor: clippy isn't complaining correctly right now, it would like this to be a `.iter()` instead of an `.into_iter()`

arckoor (approved):

src/validation.rs
mathstuf: Fixed. After that, it suggested `.to_vec()` which I've also used.

PR #474, opened 2025-12-31
Add OpenSSL crypto backend as optional feature

yongkangchen:


yongkangchen:
@Keats Once #474 lands, third-party backend support will be great.
That said, I’d like to ask whether an OpenSSL backend built on top of #474 might be worth reconsidering, given how widely OpenSSL is used and the demand from existing ecosystems.

Keats:
Not really, I'm not going to add any on top of the existing 2 for now.

PR #475, opened 2026-01-04
feat: add Jwk::from_decoding_key

eyakubovich:
Allow for constructing a Jwk from a decoding key.
This allows it to be created from a DER encoded file, for example.

This patch refactors some Jwk internals to reduce code duplication.

### DIFF
--- a/src/crypto/aws_lc/mod.rs
+++ b/src/crypto/aws_lc/mod.rs
@@ -8,7 +8,7 @@ use aws_lc_rs::{
 
 use crate::{
     Algorithm, DecodingKey, EncodingKey,
-    crypto::{CryptoProvider, JwkUtils, JwtSigner, JwtVerifier},
+    crypto::{CryptoProvider, JwtSigner, JwtVerifier, KeyUtils},
     errors::{self, Error, ErrorKind},
     jwk::{EllipticCurve, ThumbprintHash},
 };
@@ -18,7 +18,7 @@ mod eddsa;
 mod hmac;
 mod rsa;
 
-fn extract_rsa_public_key_components(key_content: &[u8]) -> errors::Result<(Vec<u8>, Vec<u8>)> {
+fn rsa_components_from_private_key(key_content: &[u8]) -> errors::Result<(Vec<u8>, Vec<u8>)> {
     let key_pair = aws_sig::RsaKeyPair::from_der(key_content)
         .map_err(|e| ErrorKind::InvalidRsaKey(e.to_string()))?;
     let public = key_pair.public_key();
@@ -26,7 +26,15 @@ fn extract_rsa_public_key_components(key_content: &[u8]) -> errors::Result<(Vec<
     Ok((components.n, components.e))
 }
 
-fn extract_ec_public_key_coordinates(
+fn rsa_components_from_public_key(key_content: &[u8]) -> errors::Result<(Vec<u8>, Vec<u8>)> {
+    let public = aws_lc_rs::rsa::PublicKey::from_der(key_content)
+        .map_err(|e| ErrorKind::InvalidRsaKey(e.to_string()))?;
+
+    let components = aws_sig::RsaPublicKeyComponents::<Vec<u8>>::from(&public);
+    Ok((components.n, components.e))
+}
+
+fn ec_components_from_private_key(
     key_content: &[u8],
     alg: Algorithm,
 ) -> errors::Result<(EllipticCurve, Vec<u8>, Vec<u8>)> {
@@ -102,9 +110,10 @@ fn new_verifier(
 pub static DEFAULT_PROVIDER: CryptoProvider = CryptoProvider {
     signer_factory: new_signer,
     verifier_factory: new_verifier,
-    jwk_utils: JwkUtils {
-        extract_rsa_public_key_components,
-        extract_ec_public_key_coordinates,
+    key_utils: KeyUtils {
+        rsa_pub_components_from_private_key: rsa_components_from_private_key,
+        rsa_pub_components_from_public_key: rsa_components_from_public_key,
+        ec_pub_components_from_private_key: ec_components_from_private_key,
         compute_digest,
     },
 };
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -10,7 +10,7 @@
 //! [`CryptoProvider`]: crate::crypto::CryptoProvider
 
 use crate::algorithms::Algorithm;
-use crate::errors::Result;
+use crate::errors::{ErrorKind, Result};
 use crate::jwk::{EllipticCurve, ThumbprintHash};
 use crate::{DecodingKey, EncodingKey};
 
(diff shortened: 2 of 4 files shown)


arckoor (changes requested):
> As an aside, why are function pointers used instead of traits?

It was what happened to work when I implemented it, I didn't explicitly choose not to use traits, there wasn't any deep logic behind it.
I didn't necessarily plan for more stuff being added in the future

src/crypto/mod.rs
```diff
@@ -123,51 +133,95 @@ See the documentation of the CryptoProvider type for more information.
...
+    };
+
+    if pub_bytes[0] != 4 {
+        return Err(ErrorKind::InvalidEcdsaKey.into());
+    }
+
+    let (x, y) = pub_bytes[1..].split_at(pub_elem_bytes);
+    Ok((curve, x.to_vec(), y.to_vec()))
+}
+
```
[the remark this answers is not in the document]
eyakubovich: I put it into the crypto backend for symmetry but also because the fact that it doesn't depend on the crypto backend is an implementation detail. From the user POV, it's a crypto related operation. Technically it's an operation on a data/file format, not a crypto algorithm but neither are the rest of the extraction functions.

If I pull it out, what module should I put into?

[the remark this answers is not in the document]
arckoor: The end user doesn't really care about this, third party providers do. If it were to remain, every other provider would need to reference it in their `CryptoProvider`, for seemingly no reason at all (also it would need to be public).

I guess you could pull it out to `jwk.rs`, but for consistency it's probably not too bad if it stays with the `CryptoProvider`
I'm not really set on that though

src/jwk.rs
```diff
@@ -627,4 +683,30 @@ mod tests {
...
+        let enc_key =
+            EncodingKey::from_ec_pem(include_bytes!("../tests/ecdsa/private_ecdsa_key.pem"))
+                .unwrap();
+        let dec_key =
+            DecodingKey::from_ec_pem(include_bytes!("../tests/ecdsa/public_ecdsa_key.pem"))
+                .unwrap();
+        let expected_jwk = Jwk::from_encoding_key(&enc_key, Algorithm::ES256).unwrap();
+        let jwk = Jwk::from_decoding_key(&dec_key, Some(Algorithm::ES256)).unwrap();
+        assert_eq!(jwk, expected_jwk);
+    }
```
arckoor: `#[cfg(feature = "use_pem")]`

arckoor:
Please rebase to latest master. A bunch of other PRs are also adding more utility functions to `JwkUtils`, so I think keeping that struct around is better in the long run, the `CryptoProvider` grows uncomfortably big otherwise. Though renaming it to something like `KeyUtils` or similar wouldn't hurt

eyakubovich:
Rebased, brought back the struct and called it `KeyUtils`.

src/crypto/mod.rs
```diff
@@ -157,17 +164,41 @@ Call CryptoProvider::install_default() before this point to select a provider ma
...
```
arckoor: This still can't be part of `KeyUtils` / the `CryptoProvider` as a whole, it's completely inaccessible from the outside of the crate. What about just making it `pub(crate)` and referring to it directly from `jwk.rs`? Should also be `ec_pub_components_...`

src/jwk.rs
```diff
@@ -487,7 +491,66 @@ impl Jwk {
                     })
                 }
                 crate::algorithms::AlgorithmFamily::Ed => {
-                    unimplemented!();
+                    unimplemented!("Edwards curve is not supported");
```
arckoor: Document this please, `from_encoding_key` should have a doc comment either way

src/jwk.rs
```diff
@@ -487,7 +491,66 @@ impl Jwk {
...
+                }
+            },
+        })
+    }
+
+    pub fn from_decoding_key(
```
arckoor, line 526: Also needs a doc comment

arckoor (changes requested):

src/jwk.rs
arckoor: Done in #514, so needs a rebase

src/crypto/mod.rs
eyakubovich: It's not meant to be accessible from outside the crate directly. It's accessible via `KeyUtils::ec_pub_components_from_public_key`. We can make it `pub(crate)` and refer directly but it seemed more consistent if it's accessible via `KeyUtils`. Right now it doesn't depend on a particular crypto impl but theoretically could. But I'm ok either way.

arckoor: It's needed to even be able to construct `KeyUtils`, so to construct a `CryptoProvider` from outside this crate you'd have to duplicate this functionality every time, and why force others to do that?

So please just remove it from `KeyUtils` altogether and make the function `pub(crate)`

src/crypto/mod.rs
```diff
@@ -157,17 +164,41 @@ Call CryptoProvider::install_default() before this point to select a provider ma
...
             },
             compute_digest: |_, _| panic!("{}", NOT_INSTALLED_OR_UNIMPLEMENTED_ERROR),
         }
     }
 }
 
+#[allow(unused)]
+fn ec_pub_components_from_public_key(
+    pub_bytes: &[u8],
+) -> errors::Result<(EllipticCurve, Vec<u8>, Vec<u8>)> {
```
arckoor: `errors::Result` should be top level imported already here

src/crypto/aws_lc/mod.rs
```diff
@@ -102,9 +110,11 @@ fn new_verifier(
...
     signer_factory: new_signer,
     verifier_factory: new_verifier,
-    jwk_utils: JwkUtils {
-        extract_rsa_public_key_components,
-        extract_ec_public_key_coordinates,
+    key_utils: KeyUtils {
+        rsa_pub_components_from_private_key: rsa_components_from_private_key,
+        rsa_pub_components_from_public_key: rsa_components_from_public_key,
+        ec_pub_components_from_private_key: ec_components_from_private_key,
+        ec_pub_components_from_public_key: crate::crypto::ec_pub_components_from_public_key,
```
arckoor: Fine to just import this

src/crypto/rust_crypto/mod.rs
```diff
@@ -108,9 +118,11 @@ fn new_verifier(
...
     signer_factory: new_signer,
     verifier_factory: new_verifier,
-    jwk_utils: JwkUtils {
-        extract_rsa_public_key_components,
-        extract_ec_public_key_coordinates,
+    key_utils: KeyUtils {
+        rsa_pub_components_from_private_key: rsa_components_from_private_key,
+        rsa_pub_components_from_public_key: rsa_components_from_public_key,
+        ec_pub_components_from_private_key: ec_components_from_private_key,
+        ec_pub_components_from_public_key: crate::crypto::ec_pub_components_from_public_key,
```
arckoor: Same here

arckoor (approved):
LGTM after the nits

PR #483, opened 2026-02-12
security: zeroize encoding and decoding keys

### LINKED ISSUE #337: Adding secrecy or zeroize for EncodingKey and DecodingKey
It probably makes sense to either add the https://crates.io/crates/secrecy or the https://crates.io/crates/zeroize crate to this crate. This way we can make sure that private and secret keys are securely erased from memory when dropped.

boxdot:
Addresses #337.

`PemEncodedKey` is still not zeroized and may contain sensistive data. Currently, it uses `simple_asn1` which does not support zeroize and also does not support parsing without taking ownership of the data.

### DIFF
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -2,6 +2,7 @@ use std::fmt::{Debug, Formatter};
 
 use base64::{Engine, engine::general_purpose::STANDARD};
 use serde::de::DeserializeOwned;
+use zeroize::{Zeroize, ZeroizeOnDrop};
 
 use crate::algorithms::AlgorithmFamily;
 use crate::crypto::{CryptoProvider, JwtVerifier};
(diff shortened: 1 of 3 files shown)


src/decoding.rs
```diff
@@ -43,7 +44,7 @@ macro_rules! expect_two {
     }};
 }
 
-#[derive(Clone)]
+#[derive(Clone, Zeroize, ZeroizeOnDrop)]
```
arckoor, line 47: Why only zeorize the `DecodingKeyKind`? Yes it's still zeroized when dropping `DecodingKey`, but deriving `Zeroize` also adds a `.zeroize()` for this struct, which can imo sometimes come in handy. Deriving it only for the key kind makes that inaccessible, since `DecodingKey.kind` is not mutably public
Just do it like you did for encoding, and skip the family

arckoor (changes requested):

arckoor:
You probably also need to rebase onto master for CI to work

src/decoding.rs
boxdot: Good point! Fixed.

arckoor:
lgtm but I'd prefer a linear history
can be squashed into a single commit

arckoor (approved):
Thanks!

PR #485, opened 2026-02-21
Fix formatting in Ed25519 key serialization

### LINKED ISSUE #484: Incorrect Thumbprint on Ed25519
The [Jwk::thumbprint()](https://github.com/Keats/jsonwebtoken/blob/master/src/jwk.rs#L534) function appears to incorrectly generate a thumbprint value using the unquoted `crv` string, which differs from how we're generating it above and how other implementations do it. It appears to have been a typo.

```rust
  EllipticCurve::Ed25519 => {
      format!(
          r#"{{crv:{},"kty":{},"x":"{}"}}"#,      // <--- Note that crv isn't double quoted.
          serde_json::to_string(&a.curve).unwrap(),
          serde_json::to_string(&a.key_type).unwrap(),
          a.x,
      )
  }
```

Earlier:

```rust
  AlgorithmParameters::EllipticCurve(a) => match a.curve {
      EllipticCurve::P256 | EllipticCurve::P384 | EllipticCurve::P521 => {
          format!(
              r#"{{"crv":{},"kty":{},"x":"{}","y":"{}"}}"#,
              serde_json::to_string(&a.curve).unwrap(),
              serde_json::to_string(&a.key_type).unwrap(),
              a.x,
              a.y,
          )
      }
      EllipticCurve::Ed25519 => panic!("EllipticCurve can't contain this curve type"),
  },
```

I will submit a PR with a fix.

scovetta:
In the thumbprint() function, the serialized format of the Ed25519 curve is missing double quotes around "crv", which would likely affect interoperability between other implementations.

Fixes #484

### DIFF
--- a/src/jwk.rs
+++ b/src/jwk.rs
@@ -531,7 +531,7 @@ impl Jwk {
                 }
                 EllipticCurve::Ed25519 => {
                     format!(
-                        r#"{{crv:{},"kty":{},"x":"{}"}}"#,
+                        r#"{{"crv":{},"kty":{},"x":"{}"}}"#,
                         serde_json::to_string(&a.curve).unwrap(),
                         serde_json::to_string(&a.key_type).unwrap(),
                         a.x,


Keats:
Thanks!

PR #487, opened 2026-03-03
fix: build wasm failed

### LINKED ISSUE #488: Cannot build with target wasm
Got this issue when try to build with target wasm
```
> cargo check --target wasm32-unknown-unknown --features rust_crypto 
    Blocking waiting for file lock on package cache
    Checking getrandom v0.2.17
error: the wasm*-unknown-unknown targets are not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support
   --> /Users/ligt/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/getrandom-0.2.17/src/lib.rs:346:9
    |
346 | /         compile_error!("the wasm*-unknown-unknown targets are not supported by \
347 | |                         default, you may need to enable the \"js\" feature. \
348 | |                         For more information see: \
349 | |                         https://docs.rs/getrandom/#webassembly-support");
    | |________________________________________________________________________^

error: could not compile `getrandom` (lib) due to 1 previous error
```

duyquang6:
Close #488

we need enable `js` feature on `getrandom` lib

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ sha2 = { version = "0.10.7", optional = true, features = ["oid"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 js-sys = "0.3"
-getrandom = "0.2"
+getrandom = { version = "0.2", features = ["js"] }
 
 [dev-dependencies]
 wasm-bindgen-test = "0.3.1"


Keats:
You can already do that manually: https://github.com/Keats/jsonwebtoken/blob/master/.github/workflows/ci.yml#L83
I would rather update the getrandom crate first

duyquang6:
> You can already do that manually: https://github.com/Keats/jsonwebtoken/blob/master/.github/workflows/ci.yml#L83 I would rather update the getrandom crate first

Should we update the `getrandom` crate first? I can help with that if needed.

Keats:
Yes, update first. I think the feature changed as well in recent versions?

duyquang6:
> Yes, update first. I think the feature changed as well in recent versions?

okay, I'll check and update it first. looks like the optional feature was changed from `js` to `wasmjs` in recent versions as well

bjorn3:
The `js` feature of `getrandom` is not meant to be enabled by libraries. The reason it exists in the first place is that wasm targets do not imply that wasm-bindgen was used or that a js environment exists in the first place. Only the final executable should enable the `js` feature if it actually uses wasm-bindgen. https://docs.rs/getrandom/latest/getrandom/#webassembly-support

PR #494, opened 2026-03-17
Make Algorithm enum non_exhaustive

### LINKED ISSUE #480: JwkUtils::compute_digest must not fail
The `compute_digest` [method](https://github.com/Keats/jsonwebtoken/blob/245858fdf3fade3ef176d49d00334dec9e9f6b0b/src/crypto/mod.rs#L147) currently does not return a `Result<>`, and thus is not allowed to error. This is not a problem for the built in providers, however third party ones may need to throw an error, and must currently [unwrap](https://github.com/arckoor/jsonwebtoken-botan/blob/1bc5139698a6e792def7a597bbd926326d556122/src/lib.rs#L109). This function should be changed to return a `Result`.

The [Algorithm](https://github.com/Keats/jsonwebtoken/blob/245858fdf3fade3ef176d49d00334dec9e9f6b0b/src/algorithms.rs#L42) enum should also be made `#[non_exhaustive]`, so that when algorithms are added in the future, third party providers continue to compile (and can panic via `unimplemented!()` instead).

Both changes are breaking.

arckoor:
Closes #480.
Given that a few other PRs are wanting to add more algorithms and that that would be a breaking change anyway I figure this can just be merged when necessary.
Also addresses `Jwk::compute_digest` not returning a `Result<>`.

### DIFF
--- a/src/jwk.rs
+++ b/src/jwk.rs
@@ -496,7 +496,7 @@ impl Jwk {
     /// Compute the thumbprint of the JWK.
     ///
     /// Per [RFC-7638](https://datatracker.ietf.org/doc/html/rfc7638)
-    pub fn thumbprint(&self, hash_function: ThumbprintHash) -> String {
+    pub fn thumbprint(&self, hash_function: ThumbprintHash) -> crate::errors::Result<String> {
         let pre = match &self.algorithm {
             AlgorithmParameters::EllipticCurve(a) => match a.curve {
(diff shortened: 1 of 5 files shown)


arckoor:
Needs an MSRV bump probably

src/jwk.rs
Keats, line 530: Since we return an error now in this function, can we not panic here and above?

arckoor: #524

PR #500, opened 2026-04-12
ci: update `pinned` to Rust 1.88

mathstuf:
Required by new `time` crates (`time`, `time-core`, `time-macros`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
                 include:
                     -   build: pinned
                         os: ubuntu-24.04
-                        rust: 1.86.0
+                        rust: 1.88.0
                     -   build: stable
                         os: ubuntu-latest
                         rust: stable


arckoor:
Thanks, but ultimately this is a MSRV problem, you can't just bump CI. Root cause is that `simple_asn1` pinned their `time` dependency recently, which forces that MSRV onto us.
I've asked whether that was really necessary [here](https://github.com/acw/simple_asn1/issues/40) (if someone sees something I don't in regards to their rfc2822 usage, please chime in).
Whatever the solution turns out to be, I'll likely address it in #498.

mathstuf:
Ok. It's just that with the way GHA works, this means that PR pipelines get no useful information until this is fixed. I understand *why* they cancel all jobs when any fails (don't waste (free) compute time), but it *can* mean more duplicate work to tackle problems one-by-one in the long run (as you run jobs over and over to just get "the next problem" instead of being able to do a single update to tackle a lot of uncovered issues).

Issue #495, opened 2026-03-21
bump rsa version to v0.10


andyof8feets:
[as per this comment](https://github.com/RustCrypto/RSA/issues/390#issuecomment-3272030410) rsa v.0.10 should at least partially address [RUSTSEC-2023-0071](https://rustsec.org/advisories/RUSTSEC-2023-0071.html).

If the change in version does not break functionality, I think it would be beneficial for this version bump

PR #502, opened 2026-04-15
refactor(rust_crypto): move RSA backend to rsa 0.10.0-rc.17

EffortlessSteven:
Addresses #495.

I kept this to the rust_crypto RSA seam.

Scope
- rust_crypto backend only
- no public API changes
- no aws_lc_rs changes
- no crate-wide signature 3 migration
- no crate-wide sha2 0.11 migration

Dependency note
- This uses the currently published crates.io 0.10 line for RSA, which is rsa 0.10.0-rc.17.

Review map
- Cargo.toml
- src/crypto/rust_crypto/rsa.rs
- src/crypto/rust_crypto/ecdsa.rs
- src/crypto/rust_crypto/mod.rs

What changed
- replace wrapper-based RSA signing with explicit prehash + low-level key operations
- keep verification on RsaPublicKey::verify(...)
- move RSA modulus/exponent handling to BoxedUint
- decouple ECDSA from rsa::pkcs8
- preserve minimal big-endian n/e extraction for DecodingKey::from_rsa_components and JWK generation
- add a wasm-target-only getrandom 0.4 alias so the existing wasm-pack commands keep working

Existing coverage already exercising the touched paths
- PEM PKCS#1 and PKCS#8 RSA round-trips
- DER round-trips
- modulus/exponent decode via DecodingKey::from_rsa_components
- JWK decode
- JWK generation from an encoding key

Validated locally
- cargo fmt --check
- cargo clippy --all-targets --features rust_crypto -- -D warnings
- cargo clippy --all-targets --features aws_lc_rs -- -D warnings
- cargo test --features rust_crypto
- cargo test --no-default-features --features rust_crypto
- cargo test --features aws_lc_rs
- cargo test --no-default-features --features aws_lc_rs
- wasm-pack test --node --features rust_crypto,getrandom/js
- wasm-pack test --node --no-default-features --features rust_crypto,getrandom/js

Non-goal
- this does not claim to fully resolve Marvin / RUSTSEC-2023-0071

### DIFF
--- a/src/crypto/rust_crypto/mod.rs
+++ b/src/crypto/rust_crypto/mod.rs
@@ -15,11 +15,21 @@ mod eddsa;
 mod hmac;
 mod rsa;
 
+fn trim_leading_zeroes(bytes: &[u8]) -> Vec<u8> {
+    match bytes.iter().position(|byte| *byte != 0) {
+        Some(first_non_zero) => bytes[first_non_zero..].to_vec(),
+        None => vec![0],
+    }
+}
+
 fn extract_rsa_public_key_components(key_content: &[u8]) -> errors::Result<(Vec<u8>, Vec<u8>)> {
     let private_key = RsaPrivateKey::from_pkcs1_der(key_content)
         .map_err(|e| ErrorKind::InvalidRsaKey(e.to_string()))?;
     let public_key = private_key.to_public_key();
-    Ok((public_key.n().to_bytes_be(), public_key.e().to_bytes_be()))
+    Ok((
+        trim_leading_zeroes(public_key.n().to_be_bytes().as_ref()),
+        trim_leading_zeroes(public_key.e().to_be_bytes().as_ref()),
+    ))
 }
 
 fn extract_ec_public_key_coordinates(
--- a/src/crypto/rust_crypto/rsa.rs
+++ b/src/crypto/rust_crypto/rsa.rs
@@ -1,6 +1,8 @@
 //! Implementations of the [`JwtSigner`] and [`JwtVerifier`] traits for the
 //! RSA family of algorithms using RustCrypto.
 
+// Use the hash types re-exported by `rsa` so this backend stays in the same
+// digest ecosystem as the 0.10 RC line without a crate-wide sha2 migration.
 use rsa::{
     BoxedUint, Pkcs1v15Sign, Pss, RsaPrivateKey, RsaPublicKey,
     pkcs1::{DecodeRsaPrivateKey, DecodeRsaPublicKey},
--- a/src/crypto/rust_crypto/rsa.rs
+++ b/src/crypto/rust_crypto/rsa.rs
@@ -1,23 +1,29 @@
 //! Implementations of the [`JwtSigner`] and [`JwtVerifier`] traits for the
 //! RSA family of algorithms using RustCrypto.
 
-use hmac::digest::FixedOutputReset;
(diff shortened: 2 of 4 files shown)


arckoor:
Let's not depend on a rc version
Once rsa 0.10 releases, cargo should be free to pull in that version, despite us specifying a lower version in the crate (since they seem to be following semver, 0.10 is a patch release, and cargo is free to resolve it. If it has relevant breaking changes, we'll resolve them then)

EffortlessSteven:
Absolutely. Leaving the RC alone makes sense.

This shook out the rust_crypto-side changes ahead of stable 0.10 and surfaced the wasm/getrandom wrinkle.

Happy to revisit once 0.10.0 is out.

src/crypto/rust_crypto/rsa.rs
```diff
@@ -53,12 +62,13 @@ fn verify_rsa<S: SignatureScheme, H: Digest + AssociatedOid>(
...
             RsaPublicKey::from_pkcs1_der(bytes)
                 .map_err(signature::Error::from_source)?
-                .verify(scheme, &digest, signature)
+                .verify(scheme, digest.as_ref(), signature)
                 .map_err(signature::Error::from_source)?;
         }
         DecodingKeyKind::RsaModulusExponent { n, e } => {
-            RsaPublicKey::new(BigUint::from_bytes_be(n), BigUint::from_bytes_be(e))?
-                .verify(scheme, &digest, signature)
+            RsaPublicKey::new(to_boxed_uint(n)?, to_boxed_uint(e)?)
```
baloo, line 69: I'd recommend the use of [BoxedUint::from_be_slice_vartime](https://docs.rs/crypto-bigint/latest/crypto_bigint/struct.BoxedUint.html#method.from_be_slice_vartime):

```suggestion
            RsaPublicKey::new(BoxedUint::from_be_slice_vartime(bytes), BoxedUint::from_be_slice_vartime(bytes))
```

NOTE(security): Now this is a variable time function (and inherently dangerous), but here this is only used on a public key and no secret is involved.

baloo:
Change looks good, but I think I would recommend to bump the other members of rustcrypto (hmac, p256, p384, ...) as this would make the dependency cleaner.
This will otherwise duplicate for the array implementation and all the crypto-common and signature traits (this is what your issues with pkcs8 stemmed from).

src/crypto/rust_crypto/mod.rs
```diff
@@ -15,11 +15,21 @@ mod eddsa;
...
+    }
+}
+
 fn extract_rsa_public_key_components(key_content: &[u8]) -> errors::Result<(Vec<u8>, Vec<u8>)> {
     let private_key = RsaPrivateKey::from_pkcs1_der(key_content)
         .map_err(|e| ErrorKind::InvalidRsaKey(e.to_string()))?;
     let public_key = private_key.to_public_key();
-    Ok((public_key.n().to_bytes_be(), public_key.e().to_bytes_be()))
+    Ok((
+        trim_leading_zeroes(public_key.n().to_be_bytes().as_ref()),
```
baloo, line 30: [BoxedUint::to_be_bytes_trimmed_vartime](https://docs.rs/crypto-bigint/latest/crypto_bigint/struct.BoxedUint.html#method.to_be_bytes_trimmed_vartime) is the function you're looking for.

NOTE(security): variable timed function, only to be used on public components.

PR #504, opened 2026-04-18
security: zeroize pem encoded key

### LINKED ISSUE #337: Adding secrecy or zeroize for EncodingKey and DecodingKey

boxdot:
`PemEncodedKey` is now zeroized and no longer contains ASN.1 parsed
data. Because `Vec<simple_asn1::ASN1Block>` owns the underlying bytes
and does not support zeroization, the full ASN.1 parsing has been
replaced by a custom partial ASN.1 implementation. This parsing is
performed only to classify the key type during construction and to
extract the key bytes on access.

This change also allows the removal of the `simple_asn1` dependency.

Closes #337 after #483 is merged.

### DIFF
--- a/src/pem/decoder.rs
+++ b/src/pem/decoder.rs
@@ -1,3 +1,5 @@
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
 use crate::errors::{ErrorKind, Result};
 
 /// Supported PEM files for EC and RSA Public and Private Keys
@@ -40,11 +42,12 @@ enum Classification {
 /// Documentation about these formats is at
 /// PKCS#1: https://tools.ietf.org/html/rfc8017
 /// PKCS#8: https://tools.ietf.org/html/rfc5958
-#[derive(Debug)]
+#[derive(Debug, ZeroizeOnDrop, Zeroize)]
 pub(crate) struct PemEncodedKey {
     content: Vec<u8>,
-    asn1: Vec<simple_asn1::ASN1Block>,
+    #[zeroize(skip)]
     pem_type: PemType,
+    #[zeroize(skip)]
     standard: Standard,
 }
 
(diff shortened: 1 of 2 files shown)


boxdot:
I had quite some fun while implementing this, especially recalling the TLV encoding and writing a custom recursive parser.

arckoor:
This will likely take me a while to review, my ASN.1 skills are rusty at best
For at least one of the keys I'd also like to see a test that directly compares the parsed point to the expected value (i.e. you manually extract them with your cryptographic library of choice, openssl, botan, ... and compare to that)
Some comments to relevant RFCs would also be helpful
May also want to throw a fuzzer at this, though that would be something to consider in general for this library
As a convenient side benefit, by no longer depending on `simple_asn1` it would allow dropping the MSRV back down :thinking:

boxdot:
Let me 
* [x] add comments about parsing (e.g. RFCs), and
* [ ] make tests less fuzzy

src/pem/decoder.rs
```diff
@@ -173,68 +152,211 @@ impl PemEncodedKey {
...
+        }
+        bytes = &bytes[1..]; // final tag byte (high bit clear)
+        0xFF // Sentinel value for any long-form tag
+    } else {
+        // Short form single-byte tag
+        first
+    };
+
+    // See <https://en.wikipedia.org/wiki/X.690#Length_octets> for length encoding
+    let len = bytes[0];
```
arckoor, line 280: Out of bounds:
```rs
let _ = read_tlv(&mut [0x1f, 0]);

// with a certificate:
let data = "
-----BEGIN PUBLIC KEY-----
AA==
-----END PUBLIC KEY-----";
let _ = DecodingKey::from_rsa_pem(data.as_bytes());
```
Add a regression test as well please

arckoor (changes requested):
I haven't reviewed the logic yet, but I fuzzed it a bit

arckoor:
I think we'll just stash this for the time being. I'm still sort of interested in it, but other PRs are again drawing in `simple_asn1`, and I don't want to force everyone to write their own ASN.1 parser, both for the complexity and the reviewing headaches. It's probably worth trying to get zeroize support into `simple_asn1` itself, likely feature gated. Happy to pass this off to any volunteers, else I'll take a look myself some time

boxdot:
Sorry, I got distracted and this PR got stale.

PR #505, opened 2026-04-18
Add fuzzers to CI

arckoor:


arckoor:
cc @boxdot something like this
realistically this should probably go into a nightly run, and there's probably a cleaner way to do it, but it should be good enough for you to either use locally or temporarily rebase on top of

Issue #244, opened 2022-03-08
Support ED25519


rgarrigue:
Hello

This is a feature request to support ed25519. The reason is explained there https://github.com/mike-engel/jwt-cli/issues/173. Sadly I'm not able to make the related PR, because I know nothing about rust.

Best regards,

Keats:
I wouldn't be against it but someone would have to submit the PR.

roy-work:
Is this feature supported? It seems to be in the library (e.g., [`from_ed_der`](https://docs.rs/jsonwebtoken/latest/jsonwebtoken/struct.EncodingKey.html#method.from_ed_der)), but attempts to use it don't seem to round-trip. (Verification always fails.)

Keats:
It's not supported

Cubxity:
Would be nice to see support for Ed25519 as the other curves [may have some weaknesses](https://safecurves.cr.yp.to/)

GunnarMorrigan:
Reading a pem ed25519 key definitely does not work.

tokarevart:
It seems like [`DecodingKey::from_ed_der`](https://docs.rs/jsonwebtoken/latest/jsonwebtoken/struct.DecodingKey.html#method.from_ed_der) function actually expects raw 32 bytes public key, which ring's [`Ed25519KeyPair::public_key`](https://docs.rs/ring/latest/ring/signature/struct.Ed25519KeyPair.html#method.public_key) returns. 
Also running `openssl asn1parse -inform der -in file-with-pubkey` shows again that these 32 bytes are not DER encoded. I've also additionally checked it using [`ed25519-dalek`](https://github.com/dalek-cryptography/curve25519-dalek/tree/main/ed25519-dalek) crate.
So essentially it seems `DecodingKey::from_ed_der` name is just misleading and `DecodingKey` doesn't yet support DER encoded ed25519 public keys. Maybe it would be a good idea to add a new method like `DecodingKey::from_ed` behaving exactly the same as existing `DecodingKey::from_ed_der`, but specified in the descrition that it expects raw 32 bytes public key. 
What to do with existing misleading name of `DecodingKey::from_ed_der` I'm not sure, changing implementation would be a (potentially) quite annoying breaking change.

tokarevart:
Regarding PEM format I had no issues while testing just now.

PR #506, opened 2026-04-28
Feat: Added support for JWK decoding Ed25519 keys

richy1623:
Summary: Added decoding JWK support for Ed25519 keys

Changes:
- Implemented the `unimplemented` `Jwk::from_encoding_key` when `Algorithm::EdDSA`
- Added to the JWK utils to support fetching Ed25519 
- Updated both backends to support extracting the `x` parameter (https://datatracker.ietf.org/doc/html/rfc8037#section-2)
- Added a new `EllipticCurve` type `Ed448` for Ed448 curves
- Updated the decoder to allow decoding PEMs of type Ed448
- Added tests for all changes

Solves: https://github.com/Keats/jsonwebtoken/issues/244


src/crypto/aws_lc/eddsa.rs
```diff
@@ -44,6 +44,28 @@ impl EdDSAVerifier {
...
+            Ed25519KeyPair::from_pkcs8(encoding_key.inner())
+                .map_err(|_| ErrorKind::InvalidEddsaKey)?
+                .public_key()
+                .as_ref(),
+        )))
+    }
+
+    pub(crate) fn get_pub_key_bytes(&self) -> Vec<u8> {
+        self.0.as_bytes().to_vec()
+    }
```
arckoor: This I'm not a fan of, all the other jwk utility functions manage their parameter extraction internally without relying on the signers. This unnecessarily couples the Signer to JWK functions

src/crypto/mod.rs
```diff
@@ -143,6 +143,9 @@ pub struct JwkUtils {
     #[allow(clippy::type_complexity)]
     pub extract_ec_public_key_coordinates:
         fn(&[u8], Algorithm) -> Result<(EllipticCurve, Vec<u8>, Vec<u8>)>,
+    /// Given a DER encoded private key and the curve type, extract the ED public key component (x)
+    #[allow(clippy::type_complexity)]
```
arckoor: The allow doesn't seem to be necessary

src/jwk.rs
```diff
@@ -487,7 +491,25 @@ impl Jwk {
                     })
                 }
                 crate::algorithms::AlgorithmFamily::Ed => {
-                    unimplemented!();
+                    // Get the curve type based off the encoding key length
+                    let curve_type: EllipticCurve = match key.inner().len() {
+                        48 => Ok(EllipticCurve::Ed25519),
+                        73 => Ok(EllipticCurve::Ed448),
+                        _ => Err(Error::from(ErrorKind::InvalidEddsaKey)),
```
arckoor, line 521: I don't understand the point of classifying `Ed448` curves if the provider just says "I don't support this" anyways. I.e. why have the extra enum value and detection logic if this could be a simple
```rs
match key.inner().len() {
    48 => Ok(...),
    _ => Err(...),
}
```
I haven't worked with 448 before, why is this one checking for length of 73 when `from_ed25519_encoding_key` is checking for 57?

tests/eddsa/mod.rs
```diff
@@ -138,3 +138,64 @@ fn ed_jwk() {
...
+#[wasm_bindgen_test]
+#[should_panic]
+fn ed_jwk_from_ed448_key() {
+    use jsonwebtoken::jwk::Jwk;
+
+    let privkey_pem = include_bytes!("private_ed448_key.pem");
+    let encoding_key = EncodingKey::from_ed_pem(privkey_pem).unwrap();
+
+    // Unimplemented so should panic
+    Jwk::from_encoding_key(&encoding_key, Algorithm::EdDSA).unwrap();
```
arckoor: Check the error type, e.g. `assert!(Jwk::from_encoding_key(...).is_err_and(|e| *e.kind() == ...);`

arckoor (changes requested):

tests/eddsa/mod.rs
richy1623: I have updated the extraction to return a `ErrorKind::UnsupportedAlgorithm` instead of using `unimplemented!(...)`

src/jwk.rs
richy1623: The mismatch was actually a mistake, 57 bytes is the length of the key itself, 73 bytes is the length of the key including the ANS.1 header.

I have however, improved the checking logic to rather use OIDs instead of length checks which is more reliable. The logic was largely lifted from the `pem/decoder#classify_pem` method where it uses ans1 headers to determine algorithm types.

src/jwk.rs
```diff
@@ -334,6 +337,7 @@ pub enum EllipticCurveKeyType {
 /// Type of cryptographic curve used by a key. This is defined in
 /// [RFC 7518 #7.6](https://tools.ietf.org/html/rfc7518#section-7.6)
 #[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize, Hash)]
+#[non_exhaustive]
```
richy1623, line 337: Following https://github.com/Keats/jsonwebtoken/commit/fa89018753f61d128fc79e7cc6f6721b2d953816 the EllipticCurve's should also be non-exhaustive

src/jwk.rs
richy1623: As for why I would prefer to include `Ed448` in the check is they are valid Ed algorithms, so to provide a clear error code (`ErrorKind::UnsupportedAlgorithm`) we need to first identify it as a valid Ed algorithm, then down the line we can state it is unsupported, rather than the catch all `InvalidAlgorithm/InvalidEddsaKey`.

richy1623:
Thanks for the review @arckoor! I have implemented the requested changes if I could please get a re-review.

src/crypto/rust_crypto/mod.rs
```diff
@@ -3,6 +3,7 @@ use ::rsa::{
     pkcs1::{DecodeRsaPrivateKey, DecodeRsaPublicKey},
     traits::PublicKeyParts,
 };
+use ed25519_dalek::SigningKey;
```
arckoor: Import as `Ed25519SigningKey`

src/jwk.rs
```diff
@@ -348,8 +352,18 @@ pub enum EllipticCurve {
...
+    #[serde(rename = "Ed448")]
+    Ed448,
 }
 
+// --- EllipticCurve Constants ---
+// Key Lengths
+// ED25519: https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.5
+pub(crate) const ED25519_PUBLIC_KEY_LENGTH: usize = 32;
+// ED448: https://datatracker.ietf.org/doc/html/rfc8032#section-5.2.5
+pub(crate) const ED448_PUBLIC_KEY_LENGTH: usize = 57;
```
arckoor: ok to inline

src/crypto/rust_crypto/eddsa.rs
```diff
@@ -35,7 +35,7 @@ impl JwtSigner for EdDSASigner {
     }
 }
 
-pub struct EdDSAVerifier(VerifyingKey);
+pub struct EdDSAVerifier(pub VerifyingKey);
```
arckoor: leftover

arckoor (changes requested):
Sorry this took so long to get back to.
Classifying Ed448 curves properly and later on rejecting them with `UnsupportedAlgorithm` is a nice idea, however doing it like this is misleading for PEM users:
```rs
let privkey_pem = include_bytes!("private_ed448_key.pem");
let encoding_key = EncodingKey::from_ed_pem(privkey_pem).unwrap(); // this succeeds

let claims = Claims {... };
let token = encode(&Header::new(Algorithm::EdDSA), &claims, &encoding_key).unwrap(); // obviously fails because it's not EdDSA, but there is also no Ed448 algorithm here :/
```
If I can construct a (PEM) key, I'd expect to be able to sign with it.
So I'd prefer if that bit were removed. Should also mean that you don't need `classify_ed_curve` at all anymore, and can just match based on key length

richy1623:
@arckoor I'm made the changes and added a test case for extracting the PEM key which now fails as expected 👍

arckoor:
Sorry, I should have been clearer, I meant to remove the 448 stuff entirely
It doesn't really hurt to have it, but including a code path that can exclusively be reached by using the `from_der` methods with an unsupported key just seems kind of silly

`ed_pub_components_from_public_key` is just entirely superfluous I think? If you inline the the length check to `from_decoding_key` as well

The 448 test is good to keep though

richy1623:
@arckoor I have removed all Ed448 references now (keeping the test). Let me know if there is anything else you would like me to change.

src/jwk.rs
```diff
@@ -574,7 +590,22 @@ impl Jwk {
...
+                    let (curve_type, x) = match &key.kind() {
+                        DecodingKeyKind::SecretOrDer(pub_bytes) => {
+                            match pub_bytes.len() {
+                                // ED25519: https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.5
+                                32 => (EllipticCurve::Ed25519, pub_bytes),
+                                _ => return Err(ErrorKind::InvalidEddsaKey.into()),
+                            }
+                        }
+                        _ => return Err(ErrorKind::InvalidKeyFormat.into()),
+                    };
```
arckoor, line 602: In #515 should be able to use `try_get_as_bytes` (just note to self so I don't forget :p)

arckoor (approved):
Thanks!

src/jwk.rs
```diff
@@ -514,15 +513,32 @@ impl Jwk {
                     })
                 }
                 AlgorithmFamily::Ed => {
-                    unimplemented!("Edwards curves are not supported");
+                    // Get the curve type based off the encoding key length
+                    // Note: here we will receive a DER key which contains a 16 byte ANS.1 header
+                    let curve_type: EllipticCurve = match key.inner().len() {
+                        // 16 byte header + 32 byte Ed25519 key
+                        48 => Ok(EllipticCurve::Ed25519),
```
Keats, line 520: do we need that check @arckoor? That's going to error for v2 eg `aws-lc-rs::Ed25519KeyPair::generate_pkcs8` (unless it's handled somewhere else/am misusing it)

arckoor: I mean yeah, but isn't this actually the job of `PemEncodedKey`? For the `from_der` methods, users are currently just required to pass the "correct" key format, but for PEM we do a fair amount of parsing. For public keys we at least try to extract the public key bytes from whatever wrapper there is, for private keys we just use the [inner content](https://github.com/Keats/jsonwebtoken/blob/e7ea8f0efd1b102342db6de34e2e43592e2c02e5/src/pem/decoder.rs#L153). E.g. this breaks completely
```rs
use aws_lc_rs::signature::Ed25519KeyPair;
use jsonwebtoken::{Algorithm, EncodingKey, jwk::Jwk};

fn main() {
    let rng = aws_lc_rs::rand::SystemRandom::new();
    let pkcs8 = Ed25519KeyPair::generate_pkcs8(&rng).unwrap();
    let pem = pem::encode(&pem::Pem::new("PRIVATE KEY", pkcs8.as_ref().to_vec()));
    let encoding_key = EncodingKey::from_ed_pem(pem.as_bytes()).unwrap();
    Jwk::from_encoding_key(&encoding_key, Algorithm::EdDSA).unwrap(); // panics, len 83
}
```
IMO `EncodingKey` and `DecodingKey` should convert any incoming key into a format that the rest of the library can use. If that is done properly, checks like this can just be omitted everywhere.
I have toyed with the idea of replacing `simple_asn1` with `der` (for zero alloc parsing, which would remove the zeroizing troubles from `PemEncodedKey` entirely) 
With that done I'd also like to refactor the `from_der` methods to actually accept proper DER encodings, so accepting V2 could be done there as well

Keats: Honestly it would be nice if people didn't have to know what they are using (is it der? is it pcks v1? maybe v2? pem?) and we just look at the format to infer the type of keys.

arckoor: Yes, `EncodingKey::new(bytes: &[u8], alg: Algorithm)` would be very neat to have I think
Would perhaps still need an extra distinction like we have now for HMAC, but for the asymmetric stuff it should definitely be possible

Keats: Yep, something like that was exactly was I was thinking

PR #511, opened 2026-05-14
fix(header): add custom oiat header

michaelcoll:
- add a custom header oiat introduced by some provider (Clerk, ...) with a dedicated type to avoid a permissive approach by using Value on extra.
- add/update test to valide the new field


Keats:
Non-standard headers should not get their own fields.

michaelcoll:
Alright, but @Keats what's your plan to address the issue ?

Keats:
Probably stop exposing extra as `pub` and have `get_extra<T: DeserializeOwned>(name: &str) -> Result<Option<T>>` or something to Header

michaelcoll:
Waiting for the fix then 😄 ! Thanks for your work !

PR #512, opened 2026-05-16
Make header extras a dedicated struct

### LINKED ISSUE #510: Flattening extra fields in Header errors on non-string extra fields
In the headers object there is a Hashmap for extra fields:

```Rust
    #[serde(flatten)]
    pub extras: HashMap<String, String>,
```

If a non-string extra field is passed as a header, such as a key exp time, then there is no way to validate the token as far as I can tell.

I think according to the spec this should be allowed as there are non string examples:
https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.11

Would it make more sense to deserialise to a serde_json::Value? Or possibly give the ability to turn off extensions to the header as a feature?

i.e.
```Rust
    #[serde(flatten)]
    pub extras: HashMap<String, serde_json::Value>,
```

arckoor:
Closes #510, #439

### DIFF
--- a/benches/jwt.rs
+++ b/benches/jwt.rs
@@ -1,7 +1,9 @@
 use criterion::{Criterion, criterion_group, criterion_main};
-use jsonwebtoken::{Algorithm, DecodingKey, EncodingKey, Header, Validation, decode, encode};
+use jsonwebtoken::{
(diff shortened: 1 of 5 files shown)


arckoor:
cc @Aiden-Ziegelaar @paulkarcher

Keats (approved):

PR #515, opened 2026-05-29
Unify encoding and decoding key APIs

arckoor:
All Signers / Verifiers except the HMAC ones manually verify the key family at construction, so do that for HMAC too.
Main motivation is that after exposing `as_bytes` in #452, the `DecodingKeyKind::RsaModulusExponent { .. } => unreachable!()` case is very much reachable by accident

### DIFF
--- a/src/crypto/aws_lc/ecdsa.rs
+++ b/src/crypto/aws_lc/ecdsa.rs
@@ -23,7 +23,7 @@ macro_rules! define_ecdsa_signer {
                 }
 
                 Ok(Self(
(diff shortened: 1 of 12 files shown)


arckoor:
@Keats I've considered renaming `EncodingKey::inner()` to `as_bytes()` as well, so that enc and dec key are as similar as possible, do you have any strong feelings about that?

Keats (approved):

Keats:
Ok to rename to as_bytes

PR #517, opened 2026-06-01
feat: add granular rust_crypto_* sub-features

sigma67:
This got a bit bigger than I'd like, but submitting this anyway in case you find it useful.

`rust_crypto` is currently an all-or-nothing bundle that pulls in seven crates (ed25519-dalek, hmac, p256, p384, rand, rsa, sha2) even when an application only needs one algorithm family. This PR adds four independent sub-features that let users opt into a family:

  rust_crypto_rsa   — RS256/384/512, PS256/384/512  (rsa + sha2 + rand)
  rust_crypto_ec    — ES256, ES384                   (p256 + p384 + sha2)
  rust_crypto_eddsa — EdDSA                          (ed25519-dalek)
  rust_crypto_hmac  — HS256/384/512                  (hmac + sha2)

`rust_crypto` effectively remains unchanged. Requesting an unsupported algorithm at runtime (one whose feature is not enabled) returns `ErrorKind::UnsupportedAlgorithm`.

Three latent issues were fixed as prerequisites, all semantically equivalent to the original behaviour:

- rsa.rs: `FixedOutputReset` was imported via `hmac::digest`. Changed to `sha2::digest`. Both re-export `digest v0.10.7` — same crate instance, same trait identity — so this is a no-op for existing builds, but removes the undeclared dependency on the hmac crate from the RSA path.

- ecdsa.rs: `DecodePrivateKey` was imported via `rsa::pkcs8`. Changed to `p256::pkcs8`. Both re-export `pkcs8 v0.10.x` at the same resolved version, so the trait identity is identical. This removes the undeclared dependency on the rsa crate from the EC path.

- hmac dep: `features = ["reset"]` is now declared explicitly. Previously the `reset` feature (which provides the `Reset` impl required by `Mac::reset()` in hmac.rs) was activated only because the rsa crate happened to pull it in transitively. The code always required it; this makes the requirement explicit and prevents a silent future breakage if rsa ever stops being a transitive enabler.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ aws-lc-rs = { version = "1.15.0", optional = true }
 
 # "rust_crypto" feature
 ed25519-dalek = { version = "2.1.1", optional = true, features = ["pkcs8"] }
-hmac = { version = "0.12.1", optional = true }
+hmac = { version = "0.12.1", optional = true, features = ["reset"] }
 p256 = { version = "0.13.2", optional = true, features = ["ecdsa"] }
 p384 = { version = "0.13.0", optional = true, features = ["ecdsa"] }
 rand = { version = "0.8.5", optional = true, features = ["std"], default-features = false }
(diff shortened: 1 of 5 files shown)


arckoor:
That's a lot of `#[cfg]`s :o
I understand the motivation, but as it is now every other `CryptoProvider` is all or nothing as well with no way to change that
I'm inclined to keep the code base simple, with all those configurations I'd want at least some tests, and that quickly explodes
The fix for the import issues I'd happily take in a separate PR though

Keats:
+1 for not having cfg all over the place

sigma67:
I figured as much, but thanks for considering.

>  The fix for the import issues I'd happily take in a separate PR though

I assume you're referring to the undeclared transitive dependencies?

arckoor:
> I assume you're referring to the undeclared transitive dependencies?

Yes, primarily that certain traits are imported from the "wrong" parent crate
The `reset` feature for `hmac` shouldn't technically be necessary since we internally do not reuse a `JwtSigner`, and as such the HMAC is consumed only once anyway, but since part of the api is public it's better to keep (and explicitly declare) it for now

PR #518, opened 2026-06-10
feat: add OpenSSL support for JWT signing and verification

iroykaufman:
This pull request adds support for using the `openssl` crate as a cryptographic backend for JWT signing and verification. It introduces a new optional `openssl` feature, implements all required cryptographic algorithms (HMAC, ECDSA, EdDSA, RSA) using OpenSSL, and integrates this provider into the crate’s feature selection logic. This allows users to choose OpenSSL as their cryptographic provider, alongside the existing `rust_crypto` and `aws_lc_rs` options.

This update provides an additional, widely used cryptographic backend option for users, improving flexibility and compatibility.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                backend: [ aws_lc_rs, rust_crypto ]
+                backend: [ aws_lc_rs, rust_crypto, openssl ]
         steps:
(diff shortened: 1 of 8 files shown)


Keats:
We don't really want to add more backends to the crate, if someone wants to use openSSL they should make a jsonwebtoken-openssl crate that impl CryptoProvider

arckoor:
[jsonwebtoken-openssl](https://crates.io/crates/jsonwebtoken-openssl) already exists

Issue #493, opened 2026-03-16
Validation::set_required_spec_claims is a footgun


bjorn3:
If you use ` set_audience` or `set_issuer`, it is reasonable to assume that this crate will validate that the `aud` or `iss` claims are set to an acceptable value. As it turns out however, this crate doesn't do this validation unless you call `set_required_spec_claims` with the claim in question included. Otherwise JWTs without these claims will be accepted just fine. While this is documented on the `aud` and `iss` fields, this is not documented on `set_audience` or `set_issuer`. In addition it is rather surprising behavior than can (and in fact does) lead to misconfigurations in the JWT validation.

Keats:
Yes I don't remember the details but we should have some internal list of params to check that can override what's defined in `set_required_spec_claims` if you call set_audience/set_issuer

PR #520, opened 2026-06-18
Make set_issuer and set_audience require `aud`, `iss` claims

mikkel3000:
Addresses #493

set_audience and set_issuer convenience functions now adds `aud` and `iss` to `required_spec_claims`.

### DIFF
--- a/src/validation.rs
+++ b/src/validation.rs
@@ -143,13 +143,17 @@ impl Validation {
 
     /// `aud` is a collection of one or more acceptable audience members
     /// The simple usage is `set_audience(&["some aud name"])`
+    /// Makes the `aud` claim required by adding to `required_spec_claims`
     pub fn set_audience<T: ToString>(&mut self, items: &[T]) {
+        self.required_spec_claims.insert("aud".to_string());
         self.aud = Some(items.iter().map(|x| x.to_string()).collect());
     }
 
     /// `iss` is a collection of one or more acceptable issuers members
     /// The simple usage is `set_issuer(&["some iss name"])`
+    /// Makes the `iss` claim required by adding to `required_spec_claims`
     pub fn set_issuer<T: ToString>(&mut self, items: &[T]) {
+        self.required_spec_claims.insert("iss".to_string());
         self.iss = Some(items.iter().map(|x| x.to_string()).collect());
     }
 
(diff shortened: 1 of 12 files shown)


src/validation.rs
```diff
@@ -818,7 +856,6 @@ mod tests {
         validation.set_required_spec_claims(&["exp", "iss"]);
         validation.leeway = 5;
         validation.set_issuer(&["iss no check"]);
-        validation.set_audience(&["iss no check"]);
```
arckoor: Could be augmented by not adding `"iss"` manually
Though then you can just remove the entire `set_required_spec_claims`

arckoor (commented):
Should be fine, without some more doc on `set_required_spec_claims` opens up the door for another footgun because that method unconditionally overwrites everything, it's not additive.
The entire interface is imo weirdly split between using methods on `Validation`, and setting some attributes directly. Thoughts @Keats?

Keats:
> The entire interface is imo weirdly split between using methods on Validation, and setting some attributes directly. Thoughts @Keats?

The API of the Validation struct is a bit confusing IMO (and has been adding things without too much thought put into it) and was always something I wanted to redo but never spent time on. It would be good to see what something designed from scratch for all the current features would look like.

src/validation.rs
mikkel3000: Okay, updated the PR.

mikkel3000:
> > The entire interface is imo weirdly split between using methods on Validation, and setting some attributes directly. Thoughts @Keats?
> 
> The API of the Validation struct is a bit confusing IMO (and has been adding things without too much thought put into it) and was always something I wanted to redo but never spent time on. It would be good to see what something designed from scratch for all the current features would look like.

Does that mean you do not want to use this change, and wait for a larger refactor of the validation API?
I agree that the API is confusing, but i think a proper re-implementation would (and should) bring multiple breaking changes.

Keats:
> I agree that the API is confusing, but i think a proper re-implementation would (and should) bring multiple breaking changes.

There is a new major version coming so it would be fine. 
I won't have time to implement it myself but would happily review it if someone else does it.

arckoor:
I'd like to try and get v11 out this month if possible, as it contains changes that others have been waiting for for quite a while
I'll try to get something ready in the next few days for this (anyone else is welcome as well if you're faster than me), but if it doesn't happen in this version it should also be fine, I'm considering another breaking version to address the whole `EncodingKey::from_der` situation, which the validation rewrite could then also be a part of

mikkel3000:
> I'd like to try and get v11 out this month if possible, as it contains changes that others have been waiting for for quite a while I'll try to get something ready in the next few days for this (anyone else is welcome as well if you're faster than me), but if it doesn't happen in this version it should also be fine, I'm considering another breaking version to address the whole `EncodingKey::from_der` situation, which the validation rewrite could then also be a part of

I will have a go at it and update this PR soon 👍

mikkel3000:
> I'd like to try and get v11 out this month if possible, as it contains changes that others have been waiting for for quite a while I'll try to get something ready in the next few days for this (anyone else is welcome as well if you're faster than me), but if it doesn't happen in this version it should also be fine, I'm considering another breaking version to address the whole `EncodingKey::from_der` situation, which the validation rewrite could then also be a part of

I have made a suggested API change, there is still polishing and documentation to be done, but the methods are mostly finished. 
If you like this style and want me to continue this implementation, I will happily do that.

arckoor:
From a quick glance direction seems fine, but I'm not sure about the builder style
I do like it and it makes sense using it for most things, but imo it encourages inline construction, while Validation is (like `EncodingKey` and `DecodingKey`) meant to be constructed once and then reused (the hashsets make it sort of expensive to repeatedly construct)
Offering a less inline friendly api by taking `&mut self` like the previous version did might push people in the right direction

mikkel3000:
> From a quick glance direction seems fine, but I'm not sure about the builder style
> 
> I do like it and it makes sense using it for most things, but imo it encourages inline construction, while Validation is (like `EncodingKey` and `DecodingKey`) meant to be constructed once and then reused (the hashsets make it sort of expensive to repeatedly construct)
> 
> Offering a less inline friendly api by taking `&mut self` like the previous version did might push people in the right direction

I get your point. I can make a refactor in a week or so. If you want to implement something else or continue this, feel free.

arckoor:
I think we'll just make it a v12 milestone, so no rush

PR #478, opened 2026-01-12
feat: add ES512 algorithm support using P-521 elliptic curve

arsenin-kitsoft:
Suggested changes for [!250](https://github.com/Keats/jsonwebtoken/issues/250)

PR #530, opened 2026-07-24
Allow deserializing unknown key type JWKs

### LINKED ISSUE #529: JwkSet deserialization fails the whole set on an unknown `kty` (RFC 7517 §5)
### Summary

`JwkSet` derives `Deserialize`, and `AlgorithmParameters` is an untagged enum with no fallback variant. A key whose `kty` (or `alg`) is not one of the known variants therefore fails deserialization of the **entire** `JwkSet`, not just that key. Per RFC 7517 section 5 an unrecognised key should be ignored, so one post-quantum key in a published JWKS breaks parsing of the classical keys alongside it.

### Reproduction

A JWKS with a classical key and one post-quantum ML-DSA key (`kty: "AKP"`, RFC 9964):

```rust
let set: jsonwebtoken::jwk::JwkSet = serde_json::from_str(mixed_jwks)?;
// Err: data did not match any variant of untagged enum AlgorithmParameters
```

Measured against 10.4.0 (fails on the unknown `kty`) and 9.3.1 (fails on the unknown `alg`). The set deserializes fine once the AKP key is removed. Fixture and probe available on request.

### Why this matters

RFC 7517 section 5: *"Implementations SHOULD ignore JWKs within a JWK Set that use `kty` values that are not understood by them."* As post-quantum migration begins (ML-DSA in Node today, Go stdlib in 1.27, RFC 9964 defines the `AKP` JWK form), issuers will publish AKP keys into key sets that jsonwebtoken consumers read. Because the whole `JwkSet` fails to deserialize, those consumers stop verifying every signature from that issuer, including classical RS256/ES256, on a key publication they do not control.

### Suggested fix

Two options, either works:

- Add a catch-all variant to `AlgorithmParameters` (for example an `Other`/`Unknown` arm, or `#[serde(other)]` where applicable) so an unknown key deserializes into an inert variant and `JwkSet::find(kid)` simply never selects it; or
- Deserialize `keys` as a `Vec` of raw values and skip members that fail to parse into a `Jwk`, keeping the recognised keys.

Either implements the RFC section 5 SHOULD without changing behaviour for recognised keys.

### Offer

Happy to share the reproduction harness (mixed classical + AKP JWKS plus a control) and to open a PR.

arckoor:
Also makes some more enums `#[non_exhaustive]`, we might add new Algorithms soon (e.g. #478).
Closes #529

### DIFF
--- a/src/decoding.rs
+++ b/src/decoding.rs
@@ -226,6 +226,7 @@ impl DecodingKey {
                     kind: DecodingKeyKind::SecretOrDer(out),
                 })
             }
+            AlgorithmParameters::Other(_) => Err(ErrorKind::UnsupportedAlgorithm.into()),
         }
     }
 
--- a/src/jwk.rs
+++ b/src/jwk.rs
@@ -3,6 +3,7 @@
 //! Most of the code in this file is taken from <https://github.com/lawliet89/biscuit> but
 //! tweaked to remove the private bits as it's not the goal for this crate currently.
 
+use std::collections::BTreeMap;
 use std::{fmt, str::FromStr};
 
 use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
@@ -149,6 +150,7 @@ impl<'de> Deserialize<'de> for KeyOperations {
 /// The algorithms of the keys
 #[allow(non_camel_case_types, clippy::upper_case_acronyms)]
 #[derive(Debug, PartialEq, Eq, Hash, Copy, Clone, Serialize, Deserialize)]
+#[non_exhaustive]
 pub enum KeyAlgorithm {
     /// HMAC using SHA-256
     HS256,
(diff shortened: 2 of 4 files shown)


src/decoding.rs
```diff
@@ -226,6 +226,7 @@ impl DecodingKey {
                     kind: DecodingKeyKind::SecretOrDer(out),
                 })
             }
+            AlgorithmParameters::Other(_) => Err(ErrorKind::UnsupportedAlgorithm.into()),
```
Keats, line 229: It would nice to have a value in `ErrorKind::UnsupportedAlgorithm` but probably not a blocker

src/jwk.rs
```diff
@@ -439,20 +441,31 @@ pub struct OctetKeyPairParameters {
...
     RSA(RSAKeyParameters),
     OctetKey(OctetKeyParameters),
     OctetKeyPair(OctetKeyPairParameters),
+    Other(OtherKeyParameters),
 }
 
 /// The function to use to hash the intermediate thumbprint data.
 #[derive(Debug, Clone, Eq, PartialEq)]
 #[allow(missing_docs)]
+#[non_exhaustive]
```
Keats, line 468: can there be other?

Keats (approved):

src/jwk.rs
arckoor: Not that I'm aware of currently, but I for one would appreciate a future capability of using SHA-3

Issue #250, opened 2022-04-27
Add ES512 support


Xuanwo:
Hi,

https://github.com/datafuselabs/databend/pull/5062 needs ES512 support from jsonwebtoken.

Is there anything special to take care of? I'm willing to submit a PR to add `ES512` support :rocket:

a1ien:
First you need add P-512 to ring https://github.com/briansmith/ring/issues/824

deepu105:
Seems like there is an active PR to add support for this in ring. https://github.com/briansmith/ring/pull/1631

sscarduzio:
Please add ES512, we are blocked too

Keats:
I think it will be available after https://github.com/Keats/jsonwebtoken/pull/410

arsenin-kitsoft:
@Keats ES512 is still needed, #410 is closed. Is there any help needed to push this?

PR #532, opened 2026-07-27
feat: add ES512 algorithm support using P-521 elliptic curve

arsenin-kitsoft:
Suggested changes for https://github.com/Keats/jsonwebtoken/issues/250

Replacement for https://github.com/Keats/jsonwebtoken/pull/478 with updated dependencies.


arsenin-kitsoft:
@Keats @arckoor Local tests are green for both feature sets, but there are extensive dep bumps beyond p* libs. Please advise if this is okay. Specifically questionable is the bump of `rsa` to an rc version of 0.10 (the other libs have dragged it).

arckoor:
`rc` versions are a no
is there no path at all in the deps to upgrade just p521? I assume it pulls a newer rand which is then too new for rsa?

src/crypto/rust_crypto/ecdsa.rs
```diff
@@ -85,3 +88,35 @@ define_ecdsa_signer!(Es384Signer, Algorithm::ES384, SigningKey384);
 
 define_ecdsa_verifier!(Es256Verifier, Algorithm::ES256, VerifyingKey256, Signature256);
 define_ecdsa_verifier!(Es384Verifier, Algorithm::ES384, VerifyingKey384, Signature384);
+
+// P-521 (ES512) has no `sign_recoverable()` support in RustCrypto, so it can't
+// go through the `define_ecdsa_signer!` macro used for ES256/ES384.
```
arckoor: Why not adjust the macro? We just discard the recovery ID do we not?

arsenin-kitsoft:
@arckoor Eeh, there's a dep mess with P* 0.14 and not only with the 521 one :thinking:
Should we wait for rc to release?
Because I don't think I'm equipped to untangle this one :crying_cat_face:

atezet:
Yeah, the `rsa` rc is very annoying. I think it should be possible to upgrade just the `pxxx` ones, but not just `p521`

src/crypto/rust_crypto/ecdsa.rs
arsenin-kitsoft: Yeah, that was easy :sweat_smile:

arsenin-kitsoft:
Had to bump getrandom to 0.4 as well to clear the wasm tests.

arsenin-kitsoft:
@atezet didn't work for me, unfortunately - still breaks with dep clash.

arsenin-kitsoft:
https://github.com/RustCrypto/RSA/issues/647 rsa 0.10 release track

arckoor:
That's very unfortunate. I can still fully review this if you'd like, or we can pick it up once all the dependencies are sorted out

arsenin-kitsoft:
@arckoor I've addressed the issues with tests and with macros that you've highlighted above. But I think it's better to wait for the next release of RSA.

atezet:
I briefly looked at it and if I'm not mistaken there's a dependency clash on `signature::{Signer, Verifier}`, right? Why are these traits necessary even if the `aws_lc_rs` backend is used? Or is the problem also with `getrandom` and wasm?

PR #535, opened 2026-08-17
Add support for ML-DSA signatures

### LINKED ISSUE #534: Support for ML-DSA signed JWTs
Hi, are there plans to implement support for ML-DSA signed JWTs? Is there a timeline or something similar?

[RFC 9964](https://www.rfc-editor.org/info/rfc9964/#name-new-jose-algorithms) specifies JOSE bindings for ML-DSA variants, and both aws-lc-rs and RustCrypto implement the relevant algorithms.

Although it is not super important to implement post-quantum signature support asap, other JWT libs (like [panva/jose](https://github.com/panva/jose) for TS) already do and eventually it will be crucial.

I'm happy to help implement this!

PhilSchmieder:
Closes #534 

## Summary

Adds ML-DSA signature support following FIPS 204 and [RFC 9964](https://www.rfc-editor.org/info/rfc9964/).

- Supports ML-DSA-44, ML-DSA-65, and ML-DSA-87.
- Implements both `aws_lc_rs` and `rust_crypto` backends.
- Adds PKCS#8 private-key and raw/SPKI public-key loading through DER and PEM APIs.
- Adds RFC 9964 `kty: "AKP"` JWK import, export, and thumbprints.
- Validates JWK algorithms and parameter-set-specific public-key lengths.
- Updates the documented algorithm list.

## Implementation Decisions

ML-DSA uses the RFC 9964 wire names (`ML-DSA-44`, `ML-DSA-65`, and `ML-DSA-87`) while retaining regular Rust variants such as `Algorithm::MLDSA44`.

Signing uses the deterministic variant of ML-DSA with an empty context. 

Variable `alg` exists in both the common JWK parameters and the AKP algorithm parameters as specified in RFC 9964. Serialization to `JwkWire` flattens them to one parameter which causes all sorts of problems when doing naively. I opted to introduce custom (de-) serialization logic that uses both locations. Serialization rejects conflicting values, and emits one `alg` member. Deserialization populates the AKP algorithm parameter and the common parameter.

## Dependencies Notes

- `aws-lc-rs` crate updated from 1.15 to 1.18 because the stable ML-DSA API is only available in 1.18.
- `ml-dsa` uses `signature` 3.x while this crate still directly uses `signature` 2.2.

---

Open to discussion on the implementation decisions and how dependencies like the 'signature' crate is handled especially since other issues / PRs mention this too.


src/decoding.rs
```diff
@@ -226,6 +264,54 @@ impl DecodingKey {
...
+
+                // The declared parameter set fixes the exact public-key length
+                // (FIPS 204). Reject keys whose `pub` does not match.
+                let decoded = b64_decode(&params.pub_)?;
+                let expected_len = match alg {
+                    Algorithm::MLDSA44 => ML_DSA_44_PUBLIC_KEY_LEN,
+                    Algorithm::MLDSA65 => ML_DSA_65_PUBLIC_KEY_LEN,
+                    Algorithm::MLDSA87 => ML_DSA_87_PUBLIC_KEY_LEN,
+                    // Unreachable: family check above guarantees an ML-DSA alg.
+                    _ => return Err(new_error(ErrorKind::InvalidAlgorithm)),
```
arckoor: Either it is unreachable, in which case it should be `unreachable!()`, or by putting wrongly formatted stuff somewhere you can reach it, in which case the comment is wrong

tests/ml_dsa/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+    Algorithm::MLDSA65,
+    "private_ml_dsa_65.der",
+    "public_ml_dsa_65.raw"
+);
+ml_dsa_der_round_trip!(
+    round_trip_der_87,
+    Algorithm::MLDSA87,
+    "private_ml_dsa_87.der",
+    "public_ml_dsa_87.raw"
+);
```
arckoor: Why a macro, just use a loop?

arckoor (commented):
Very hasty review only, but looks like the right direction.
- How were the keys for the tests generated? OpenSSL, or through rust-crypto, or something else entirely? (just curious)
- The JWK dance with serde I'm not a huge fan of, I would like to find a way to make that unnecessary

tests/ml_dsa/mod.rs
PhilSchmieder: With macros you can run the tests individually. But also I wouldn't really mind a loop.

arckoor: Makes it less readable imo, especially if the test happens to fail

PhilSchmieder:
> Very hasty review only, but looks like the right direction.
> 
>     * How were the keys for the tests generated? OpenSSL, or through rust-crypto, or something else entirely? (just curious)
> 
>     * The JWK dance with serde I'm not a huge fan of, I would like to find a way to make that unnecessary

I used OpenSSL to generate the test keys.

I also don't like the serialization stuff. I guess than can be fixed with either changing the "JwkWire" representation or the AKP params.
JwkWire could not be flattened or changed in some other way, but: 1) I don't know what other stuff breaks as a result of that; and 2) That still leaves two spots where the alg is specified and which would potentially be reconciled.
On the other hand, the RFC specifies the alg in the AKP params. Changing that would break the neat encapsulation of the parameters. This would also be annoying when future signature scheme might also rely on the AKP parameters.

I tend to keeping the alg member in the AKP parameters to be RFC compliant and to future proof.

arckoor:
The `wasm_js` feature shouldn't be enabled, at least not by us (see https://crates.io/crates/getrandom). Not critical to change / fix right now, I've been meaning to bump some of the deps anyway, which hopefully fixes things then.
As an aside, your commits don't seem to be authored by your github user, you may want to either change that or at least verify the email you're committing with :)

I'll try to do a full review sometime next week.

src/crypto/aws_lc/ml_dsa.rs
```diff
@@ -0,0 +1,90 @@
...
+                        .map_err(|_| ErrorKind::InvalidKeyFormat)?,
+                ))
+            }
+        }
+
+        impl Signer<Vec<u8>> for $name {
+            fn try_sign(&self, msg: &[u8]) -> std::result::Result<Vec<u8>, Error> {
+                let mut signature = vec![0u8; self.0.algorithm().signature_len()];
+                let len = self.0.sign(msg, &mut signature).map_err(Error::from_source)?;
+                signature.truncate(len);
```
arckoor, line 35: Why the truncate? The doc for `.signature_length` does not read like it provides a bound, but rather an exact number of bytes. If the resulting signature length is not equal to what `.sign` produces, that sounds like an issue (and should instead be handled with an `Err()`?)

src/crypto/rust_crypto/ml_dsa.rs
```diff
@@ -0,0 +1,168 @@
...
+                assert!(verify(&sig, msg, &decoding_key, $alg).unwrap());
+                // A tampered message must not verify.
+                assert!(!verify(&sig, b"tampered", &decoding_key, $alg).unwrap());
+            }
+        };
+    }
+
+    round_trip_test!(round_trip_ml_dsa_44, MlDsa44, Algorithm::MLDSA44);
+    round_trip_test!(round_trip_ml_dsa_65, MlDsa65, Algorithm::MLDSA65);
+    round_trip_test!(round_trip_ml_dsa_87, MlDsa87, Algorithm::MLDSA87);
```
arckoor, line 135: I'd still prefer no macros in tests please

src/decoding.rs
```diff
@@ -200,6 +204,41 @@ impl DecodingKey {
...
+        // The raw public key must be one of the fixed FIPS 204 sizes.
+        match decoded.len() {
+            ML_DSA_44_PUBLIC_KEY_LEN | ML_DSA_65_PUBLIC_KEY_LEN | ML_DSA_87_PUBLIC_KEY_LEN => {}
+            _ => return Err(new_error(ErrorKind::InvalidKeyFormat)),
+        }
+        Ok(DecodingKey {
+            family: AlgorithmFamily::Mldsa,
+            kind: DecodingKeyKind::SecretOrDer(decoded),
+        })
+    }
```
arckoor, line 240: Do people actually use this? For a bit of background, I'd like to refactor the `EncodingKey` and `DecodingKey` interfaces in the future and combine all the `from_*_pem` methods into a single `from_pem` and the like, and I'm not yet sure how many of the non-PEM methods will stay. If there is a legit usecase for this I don't mind it being added, but if it's just there to have it I would omit it until there is some demand for it

src/jwk.rs
```diff
@@ -257,14 +276,22 @@ impl TryFrom<KeyAlgorithm> for Algorithm {
...
 
 impl fmt::Display for KeyAlgorithm {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{:?}", self)
+        match self {
+            KeyAlgorithm::MLDSA44 => write!(f, "ML-DSA-44"),
+            KeyAlgorithm::MLDSA65 => write!(f, "ML-DSA-65"),
+            KeyAlgorithm::MLDSA87 => write!(f, "ML-DSA-87"),
+            other => write!(f, "{:?}", other),
+        }
```
arckoor: Comment here please describing why it's necessary re serde and dash notation

src/algorithms.rs
```diff
@@ -15,6 +22,8 @@ pub enum AlgorithmFamily {
     Ec,
     /// Edwards curve public key family.
     Ed,
+    /// ML-DSA public key family.
+    Mldsa,
```
arckoor, line 26: `MlDSA`? To match `EdDSA`?

src/algorithms.rs
```diff
@@ -88,11 +108,37 @@ impl FromStr for Algorithm {
...
+            Algorithm::PS512 => "PS512",
+            Algorithm::EdDSA => "EdDSA",
+            Algorithm::MLDSA44 => "ML-DSA-44",
+            Algorithm::MLDSA65 => "ML-DSA-65",
+            Algorithm::MLDSA87 => "ML-DSA-87",
+        };
+        f.write_str(s)
+    }
+}
+
```
arckoor, line 141: Either do both `fmt::Display` for `Algorithm` and `KeyAlgorithm` like this, or use `other => write!(f, "{:?}", other)` here

arckoor (changes requested):
Sorry this took so long.
I played around with the `JwkWire` thing for a while, I didn't find a nice way to get rid of it. However I would not duplicate the `alg` field between them, and instead deser the json into `JwkWire`, reassemble into `Jwk` and beforehand check if `alg` needs to be present
```rs
if let AlgorithmParameters::AlgorithmKeyPair(_) = &algorithm {
    common.key_algorithm.ok_or_else(|| de::Error::missing_field("alg"))?;
}
```

src/crypto/aws_lc/ml_dsa.rs
PhilSchmieder: Ah, you're right, I'll remove it. I think that's left over from an older version of the function I wrote and later missed.

src/algorithms.rs
PhilSchmieder: I don't mind, any way is fine. I opted for `Mldsa` to mirror the other members of `AlgorithmFamily` like `Hmac`.

src/decoding.rs
PhilSchmieder: I don't think so, people should generally use `from_jwt` or other functions. Internally `from_jwt` uses the `from_*_components` functions for other sigs, but I didn't use it for ML-DSA to check against the specific key sizes that are standardized in FIPS204 and identified by the ML-DSA variant in `alg`.
What's the best play here? I'd keep the size check since ML-DSA variants are only standardized for specific sizes not like RS256 for example.

arckoor: The size check is definitely needed yes, I was mostly talking about not including the entirety of `from_mldsa_components` in the (public) API.

src/algorithms.rs
arckoor: that's fair, let's keep `Mldsa` then

src/decoding.rs
PhilSchmieder: Simplifying the API sounds like a good idea! People shouldn't really touch the individual components of a cryptographic key. Especially not in a JWT library.

PhilSchmieder:
Hi, can you please review the changes I made when you get a chance? Thanks!

src/decoding.rs
```diff
@@ -1,10 +1,14 @@
 use std::fmt::{Debug, Formatter};
+use std::unreachable;
```
arckoor: unneeded afaik

arckoor (commented):
LGTM from a semi quick glance. The `getrandom` situation is really unfortunate though, I think we'll update the rust crypto stack as a whole and then this can be merged, see the discussion in #536

src/decoding.rs
arckoor: Let's just get rid of this entirely for now

PhilSchmieder: Good call :+1:

arckoor:
@PhilSchmieder I will give this a look again when it's time to merge it, which won't be at least until the next major and until the rust crypto people release rsa v10 so we can bump that stack. If you want to prepare for that you can remove your getrandom changes from the `Cargo.toml` so that hopefully when #536 or something similar is merged it'll cause no / fewer conflicts.

If you're very bored you can have a look at the [jsonwebtoken-providers](https://github.com/arckoor/jsonwebtoken-providers) repo, I'd take an accompanying PR there, though subject to the same "won't be merged until ..." restrictions, so might be better to tackle that once this is actually merged.

PR #537, opened 2026-09-10
feat: adding support for Unsecured JWT without signature algorithm type "none"

douggynix:
# Unsecure JWTs Decoding
Here I add support for decoding JWT with header algorithm set to value "none". The RFC has it specified. But JSONWeb token library didn't support decoding such JWT and would always fail. I added None to the list of variants of Algorithm enums to be able to support unsecure jwt decoding.

Here is the RFC and the section of the pages where this is mentioned as a spec to JWT which jsonwebtoken lib didn't support.
https://datatracker.ietf.org/doc/html/rfc7519#page-12
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "jsonwebtoken"
-version = "11.0.0"
+version = "11.0.1"
 authors = ["Vincent Prouillet <hello@vincentprouillet.com>"]
 license = "MIT"
 readme = "README.md"
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -15,6 +15,8 @@ pub enum AlgorithmFamily {
     Ec,
     /// Edwards curve public key family.
     Ed,
+    /// For Insecured JWT with "alg" : "none"
+    None,
 }
 
 impl AlgorithmFamily {
@@ -32,6 +34,7 @@ impl AlgorithmFamily {
             ],
             Self::Ec => &[Algorithm::ES256, Algorithm::ES384],
             Self::Ed => &[Algorithm::EdDSA],
+            Self::None => &[Algorithm::None],
         }
     }
 }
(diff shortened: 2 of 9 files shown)


Keats:
It's not supported on purpose. Use https://docs.rs/jsonwebtoken/latest/jsonwebtoken/dangerous/fn.insecure_decode.html if you need to handle the "none" case.

douggynix:
> It's not supported on purpose. Use https://docs.rs/jsonwebtoken/latest/jsonwebtoken/dangerous/fn.insecure_decode.html if you need to handle the "none" case.

It is still an issue. The issue is with insecure_decode. The serializer always failed when the header field has **alg** with value **none**  as below
```json
{
  "alg": "none",
  "typ": "JWT"
}
```

Even though you skip validations for with insecure_decode functions, the deserialization is always failing as the enum you use for Algorithm values doesn't have "none" in its list . I guess you may reconsider reopen up this PR as this is a real issue. 
For a simple test try to decode that legit Unsecure JWT with header type alg = none as below.  
Here is one example to try out. I even have a unit test case from this PR to handle particularly that case: I am going to open up an issue for a more understanding of the deserialization problem 
```rust
    let token = "eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.eyJzdWIiOiJqb2huQGVtYWlsLmNvbSIsIm5hbWUiOiJKb2huIERvZSIsImlhdCI6MTUxNjIzOTAyMiwiZXhwIjoxNTE2MjM5MDQwLCJhdWQiOlsianNvbndlYnRva2VudGVzdCJdfQ.";

    let TokenData { header, claims } = insecure_decode::<Claims>(token).unwrap();

    assert_eq!(Some("JWT".to_string()), header.typ);

    assert_eq!(vec!["jsonwebtokentest"], claims.aud);
    assert_eq!("john@email.com", claims.sub);
    assert_eq!(1516239022, claims.iat);
    assert_eq!(1516239040, claims.exp);
```

> called `Result::unwrap()` on an `Err` value: Error(Json(Error("unknown variant `none`, expected one of `HS256`, `HS384`, `HS512`, `ES256`, `ES384`, `RS256`, `RS384`, `RS512`, `PS256`, `PS384`, `PS512`, `EdDSA`", line: 1, column: 13)))
stack backtrace:

